PR #6, opened 2022-09-22
Pulsar adjustments

mauricioszabo:
Basically, re-adjusting things here that are still pointing to "atom home" to start searching on "pulsar home"

### DIFF
--- a/src/rebuild-module-cache.coffee
+++ b/src/rebuild-module-cache.coffee
@@ -20,7 +20,7 @@ class RebuildModuleCache extends Command
       Usage: apm rebuild-module-cache
 
       Rebuild the module cache for all the packages installed to
-      ~/.range/packages
+      ~/.pulsar/packages
 
       You can see the state of the module cache for a package by looking
       at the _atomModuleCache property in the package's package.json file.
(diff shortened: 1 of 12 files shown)


src/rebuild-module-cache.coffee
```diff
@@ -20,7 +20,7 @@ class RebuildModuleCache extends Command
       Usage: apm rebuild-module-cache
 
       Rebuild the module cache for all the packages installed to
-      ~/.atom/packages
+      ~/.range/packages
```
confused-Techie, line 23:
```suggestion
      ~/.pulsar/packages
```

This seems like the value we would expect here?

confused-Techie (changes requested):
Added a single comment I think got missed. Otherwise this looks great.

src/rebuild-module-cache.coffee
mauricioszabo: Great catch

confused-Techie (approved):
Thanks for being so fast on my suggestion. This looks like a great contribution to me

confused-Techie:
Also I believe the reason that test is failing, is because its expecting a different return, which you've changed to `The Pulsar or Atom engine range in the package.json file is invalid:` because that's what's accurate, but the test expects `The Atom engine range in the package.json file is invalid:`. Should be able to just update the string it expects in the test to this

confused-Techie:
Yup, should just change the string expected in `spec/publish-spec.coffee` Line 77

PR #8, opened 2022-09-26
Various Revisions

confused-Techie:
This PR covers several points of concern for the repo.

While this is still in progress the already completed/intended to be complete areas are:

* Decaffeinate everything, from specs to src.
  - A note on decaffeination, while it may not be the prettiest code, it was decaffeinated with the tools that automatically did it during building. Meaning it should be nearly the exact some code that has been shipped so far. Exempt in some key areas:
    - Obvious uses of `var` are being replaced with the appropriate `let` or `const`
    - Additionally any instances of internal Express Servers being used are being rewritten, more on that later.
    - Pointless returns are being removed as needed.
* Updates some dependencies:
  - `express` 4.17.1 => 4.18.1
  - `node-gyp` 5.1.1 => 9.1.0
* Express Server code is being rewritten with modern conventions and syntax, to be more accessible to those newer with the platform, so that their docs match what we have.
* All Coffee related dependencies and scripts have been removed
* Scripts have been modified to support Windows systems that contain a path in the name. This resolves PR #4, while ignored whatever formatting error that happened on that PR.

Some things that should still be high priority to complete even if not present in this PR.

* We should get `git-utils` to rely on our own fork. Although our fork doesn't seem to contain the tagged version referenced in our dependency. And downgrading to `5.7.0` produces many errors, and will likely have to be revisited.
* Ideally we can move away from custom implementations of Jasmine like we have here `jasmine-focused`, which relies on a specific commit of `kevinsawicki/jasmine-node` which itself is forked from `mhevery/jasmine-node` meanwhile `jasmine` supports node just fine at modern versions and could likely save us some headache of finding documentation. But unfortunately our current tests rely on some features that no longer or have never existed on normal `jasmine`, so those will have to get significant rewrites to work.

Otherwise this PR will remain a draft until work is done on properly decaffeinating and properly rewriting the Express code. 

Feel free to add more here, especially around dependencies, as I'd like to get everything updated that we can without to many mysterious test failures, and switch over to all of our forks for dependencies as much as possible.


DeeDeeG:
Just a brief comment, the `node-gyp` dependency being separate was only to resolve a quirk during testing -- depending on how how deeply or shallowly nested npm decides to hoist the `node-gyp` module under `node_modules` folder, its path can change. Whereas there is one spec that tries to copy `node-gyp` from its herd-coded path under `node_modules`.

Amin decided to pin a separate copy of node-gyp so it would always be hoisted to top-level. This was done over at the atom-community/apm repo.

Thes copy won't actually be used in real-world usage, only the copy required by npm will actually be used, so I recommend keeping it the same version so it will dedupe with npm's dependency of node-gyp.

If we _really_ want newer node-gyp, in a more meaningful way, I have a branch over at atom-community/apm that bumps to npm 7, and here is a PR someone submitted to that repo to get ready for npm 8: https://github.com/atom-community/apm/pull/123

confused-Techie:
This PR is far to out of date to even be considered from merging.

But the efforts made within are still perfectly valid, and encouraged to be retake to anyone who has time. Eventually I myself would really like to take another shot at this, but doing so properly:

* Do not decaf spec and source at the same time
* Breaking up each item into singular PRs to properly and easily review
* Have codacy configured properly to not ruin any chance at successful PR comments (This one is already done)

PR #9, opened 2022-10-04
Change to pnpm so this won't fail pulsar in building

kaosine:
Needed this converted too...problem is it keeps having errors, but posting it anyway just to see if it's my own machine w/toolbox. Can't seem to get vagrant working anyway...


Spiker985:
I'm just gonna close this, since it seems we're either going to stick with npm or yarn

PR #12, opened 2022-10-22
removed atom.io references

Sertonix:


mauricioszabo:
@Sertonix there's a conflict that should be easy to solve. I could do it, but it'll commit directly to `master`, so can you fix it? Then it should be good to go :)

PR #11, opened 2022-10-09
ci: Use 'windows-2019' image for Windows tests (makes CI passing)

bongnv:
The fix is copied from https://github.com/atom/github/pull/2758:

> This gives us an older version of Visual Studio compatible with apm's old copy of node-gyp, so CI can proceed past installing dependencies/not get stuck, on the Windows jobs.
>
> (context: ppm is still built around npm 6 --> node-gyp 5.x. Visual Studio 2022 support was only added in [node-gyp 8.4.0](https://github.com/nodejs/node-gyp/blob/master/CHANGELOG.md#840-2021-11-05). So we need an older Visual Studio to accommodate ppm's old copy of node-gyp. the 'windows-2019' image has the older Visual Studio we're looking for.)

PR #39, opened 2022-11-17
Really Fix Windows CI

DeeDeeG:
**Note: I'm not sure why I can't write concise technical stuff lately, but this is a few very minor changes to make CI start passing on Windows again.**

(Feel free to not read all that I've written here, because it is honestly a lot to read, and kind of an unnecessary level of detail. A bit of a brain dump. It probably takes less time to read the entire diff and commit messages than this massive PR body.)

This does a couple of things, essentially just to play whack-a-mole with fixing install errors in CI:
- Stops trying to update the global copy of npm in the CI instance, as this was failing on the windows-2019 image with Node 14 only, for some unknown reason???
  - File system permission issues when deleting the old copy of npm???
- Once again fixes the recurring issue where the package **`jasmine-node`** is specified (as an indirect dependency under the `jasmine-focused` package...) using an **_GIT+SSH://_** URL schema, which fails extremely often in various environments (especially CI), and confuses npm tooling extremely often and fatally, with poor and cryptic error messages being generated that are not very helpful.
  - if you see `ssh://` in package-lock.json, it should be replaced with the equivalent `https://` URL _**ASAP**_ to prevent this sort of issue from happening again in the future. Sadly, any large amount of lockfile churn will tend to reintroduce the `ssh://` stuff again...
  - We should really fork `jasmine-focused` just to update its package.json to stop specifying its dependency on `jasmine-node` with a weird `git+https://` schema URL... ([here](https://github.com/atom/jasmine-focused/blob/f2fce423323d637e1b85a960240c1c974718787c/package.json#L38).)
  - This was causing issues with npm v6.x bundled with Node 14.x
- Bonus: Set the `check-latest: true` option for actions/setup-node, so we always get the latest patch versions of the Node LTS versions we specify. These Node patch-level updates also occasionally include patch-level bumps of Node's bundled npm as well, occasionally.
- Also not strictly necessary but bonus: Use the lockfile v1 format of `package-lock.json` used by npm 6.x. I would humbly suggest we decide on **_either_** lockfile v1 (npm 6.x) or lockfile v2 (npm 7.x or 8.x)... or even lockfile v3 (npm 9.x! which I recently learned exists... Sigh.)

This PR is a follow-up to https://github.com/pulsar-edit/ppm/pull/11.


confused-Techie (approved):
This all sounds awesome to me!

But curious if we could end up just using another testing framework other than jasmine faster than maintaining a fork of it. But we will have to see. Not opposed to forking it, and letting our lives just be a tiny bit easier though

DeeDeeG:
Thanks for the review!

And by the way, with forking `jasmine-focused`, I mean doing so with the sole ambition of updating the way `jasmine-node` is specified in `package.json` from `git+https://` to just `https://`. Should be a one-line change of removing four characters, if I'm picturing it right, just off the top of my head. This has come up as an issue I don't know how many times over at upstream and the atom-community fork.

Good news: the absolute newest npm versions appear to cope with it the `git+https://` stuff just fine, so maybe it's just npm 6.x that's affected. And npm 6 won't be supported forever, so this won't be an issue forever even with no forking. But uh, getting kinda off on this tangent. I can propose it in another PR if I get around to it.

Anyhow... Merging this!

PR #43, opened 2022-12-03
src: Update Electron header download URL

DeeDeeG:
The old atom.io Electron headers URL is going down, per the updated sunset announcement.

Likewise, the old gh-contractor-zcbenz Amazon AWS/S3 bucket is no-longer guaranteed to be active, and may go away at any time.

(See: https://github.blog/2022-06-08-sunsetting-atom/#if-im-using-atom-what-changes-can-i-expect-after-the-sunset and https://www.electronjs.org/blog/s3-bucket-change)

So, this commit updates our package manager to get the Electron headers from the new 'artifacts.electronjs.org/headers/dist' location.

This is the correct, official place to get the Electron headers, according to the blog post above at electronjs.org.

Ensures we can continue to build native C/C++ code for certain Atom/Pulsar packages that use it.

(For example: several of the language packages have C/C++ addon code.)

### DIFF
--- a/src/apm.coffee
+++ b/src/apm.coffee
@@ -67,7 +67,7 @@ module.exports =
     process.env.ATOM_REPOS_HOME ? path.join(@getHomeDirectory(), 'github')
 
   getElectronUrl: ->
-    process.env.ATOM_ELECTRON_URL ? 'https://atom.io/download/electron'
+    process.env.ATOM_ELECTRON_URL ? 'https://artifacts.electronjs.org/headers/dist'
 
   getAtomPackagesUrl: ->
     process.env.ATOM_PACKAGES_URL ? "#{@getAtomApiUrl()}/packages"


confused-Techie (approved):
Sorry for lack of approval, lets get this merged

DeeDeeG:
Copy-pasting from Discord, testing notes:

> I tested, and it is working for me locally. (Also tested with other deliberately incorrect URLs to confirm that breaks it, to disprove the "null hypothesis" that my change wasn't doing anything. BTW, you can do just https://electronjs.org/headers and it will redirect, but that causes unnecessary HTTP 301 and HTTP 307 redirect messages to go back and forth across the wire and redirect to the actual final location, hurting perf and wasting some bits of network usage. So I went with this direct URL that responds with HTTP 200, no redirects needed. ✔️ )

Best way I know of to test this is to use the changed code to install a Pulsar package that includes a native C/C++ module. For example, I run `ppm install --verbose language-carp` since it is a small package that builds a native C/C++ add-on as part of its install process.

The `--verbose` flag includes a lot of info, including every network request needed to fetch Electron headers. You may need to delete existing headers cached under `~/.pulsar/.node-gyp/.cache/node-gyp`, because if you've already downloaded these once, `node-gyp` skips re-downloading. They are cached for efficient re-use. So make sure to delete those between test runs, or this URL is irrelevant since no actual requests for the header files go out over the wire for this URL, in case of an existing cached header tarball.

PR #48, opened 2023-01-01
Fix .com links, pulsar rebranding and rebranding readme

Daeraxa:
Mostly this just fixes a few links:

- Org .github CONTRIBUTING.md link
- Update of `.com` to `.dev` links

However I also did a quick rebrand of the readme at the same time and I'm not sure if this is technically correct or not.

Elsewhere in the documentation we almost exclusively use `pulsar -p` as the "new" alias of `apm`.

However here it makes sense to keep the "direct" style of command in my mind. However this is where there is an issue because the binary itself is still `apm`.

So what I've done is change all `apm` instances to `ppm` in the style of https://github.com/pulsar-edit/pulsar/pull/273 this PR that is meant to add `ppm` to the PATH.

However this is also still correct if we used `apm` if we were talking about executing the binary directly so I don't know what the correct answer is here - feel free to make suggestions as to what we should do here. There are a *lot* of ppm command outputs that still talk about `apm` commands so whatever we choose here is also what we should be rolling out to the rest in a future pr.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,23 +1,22 @@
-# apm - Atom Package Manager
+# ppm - Pulsar Package Manager
 
-![Build Status](https://github.com/atom-ide-community/apm/workflows/CI/badge.svg)
-[![Dependency Status](https://david-dm.org/atom/apm.svg)](https://david-dm.org/atom/apm)
+Discover and install Pulsar packages powered by [pulsar-edit.dev](https://web.pulsar-edit.dev).
 
-Discover and install Atom packages powered by [atom.io](https://atom.io)
+ppm is bundled with the `pulsar` binaries so any ppm command can also be run with `pulsar -p` or `pulsar --package`.
 
-You can configure apm by using the `apm config` command line option (recommended) or by manually editing the `~/.atom/.apmrc` file as per the [npm config](https://docs.npmjs.com/misc/config).
+You can configure ppm by using the `ppm config` command line option (recommended) or by manually editing the `~/.pulsar/.apmrc` file as per the [npm config](https://docs.npmjs.com/misc/config).
 
 ## Relation to npm
 
-apm bundles [npm](https://github.com/npm/npm) with it and spawns `npm` processes to install Atom packages. The major difference is that `apm` sets multiple command line arguments to `npm` to ensure that native modules are built against Chromium's v8 headers instead of node's v8 headers.
+ppm bundles [npm](https://github.com/npm/npm) with it and spawns `npm` processes to install Pulsar packages. The major difference is that `ppm` sets multiple command line arguments to `npm` to ensure that native modules are built against Chromium's v8 headers instead of node's v8 headers.
 
-The other major difference is that Atom packages are installed to `~/.atom/packages` instead of a local `node_modules` folder and Atom packages are published to and installed from GitHub repositories instead of [npmjs.com](https://www.npmjs.com/)
+The other major difference is that Pulasr packages are installed to `~/.pulsar/packages` instead of a local `node_modules` folder and Pulsar packages are published to and installed from GitHub repositories instead of [npmjs.com](https://www.npmjs.com/)
 
-Therefore you can think of `apm` as a simple `npm` wrapper that builds on top of the many strengths of `npm` but is customized and optimized to be used for Atom packages.
+Therefore you can think of `ppm` as a simple `npm` wrapper that builds on top of the many strengths of `npm` but is customized and optimized to be used for Pulsar packages.
 
 ## Installing
 
-`apm` is bundled and installed automatically with Atom. You can run the _Atom > Install Shell Commands_ menu option to install it again if you aren't able to run it from a terminal (macOS only).
+`ppm` is bundled and installed automatically with Pulsar. You can run the _Pulsar > Install Shell Commands_ menu option to install it again if you aren't able to run it from a terminal (macOS only).
 
 ## Building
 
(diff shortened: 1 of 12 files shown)


src/auth.coffee
```diff
@@ -9,7 +9,7 @@ catch error
   else
     throw error
 
-tokenName = 'Atom.io API Token'
+tokenName = 'pulsar-edit.dev API Token'
```
DeeDeeG, line 12: ⚠️ This and the next one will probably require users logging back into the Pulsar package API, or else copying over their token they added when this used/overwrote the Atom API token. We may get support requests over this.

I think it is the right thing to do, to not clobber the existing Atom token, but something to be aware of.

src/auth.coffee
```diff
@@ -36,4 +36,4 @@ module.exports =
   #
   # token - A string token to save.
   saveToken: (token) ->
-    keytar.setPassword(tokenName, 'atom.io', token)
+    keytar.setPassword(tokenName, 'pulsar-edit.dev', token)
```
DeeDeeG, line 39: ⚠️ Ditto, this will also probably cause users to have to re-auth to the Pulsar API, or find and copy their existing token. Likely to cause some support requests. But I again lean toward thinking this is the right thing to do. Not clobber (over-write) the existing Atom token users might have, and make our token findable searching for "Pulsar" instead of "Atom".

src/install.coffee
```diff
@@ -40,7 +40,7 @@ class Install extends Command
              apm install --packages-file my-packages.txt
              apm i (with any of the previous argument usage)
 
-      Install the given Atom package to ~/.atom/packages/<package_name>.
+      Install the given Atom package to ~/.pulsar/packages/<package_name>.
```
DeeDeeG, line 43:
```suggestion
      Install the given Pulsar package to ~/.pulsar/packages/<package_name>.
```

We can update this from "Atom" to "Pulsar"?

DeeDeeG (approved):
Thank you so much for doing this!! (And sorry it took so long to review!!)

I looked over every change, it all looks good other than a few comments I pointed out, these are not blockers IMO, just things I wanted to point out, a type on one line, another spot to change "Atom" to "Pulsar".

I specifically reviewed every URL in the changes, and they looked correct and semantically/IRL proper 1:1 replacements of the atom.io or pulsar-edit**.com** links they are replacing. Thank you for taking a close eye to the details and getting those all lined up just right pointing to the right endpoints/pages.

\[ ⚠️ Note: This PR changes the name/domain of the package API token ppm stores, so it no-longer overwrites or re-uses the atom.io API token users may already have. I think this is the right thing to do, but be aware it will cause users to need to log in again, even if they didn't have to before this change, or even if they've already logged in to the Pulsar package API once bfore (in which case they can locate and copy their old token). \]

\[ 📖 For users to find their old token, they can use the system credential manager as described here: https://github.com/atom/github/issues/2686#issuecomment-851136891. Probably using [`seahorse`](https://packages.ubuntu.com/jammy/seahorse) on Ubuntu, [Credential Manager](https://support.microsoft.com/windows/accessing-credential-manager-1b5c916a-6a16-889f-8581-fc16e8165ac0) on Windows, and [Keychain Access](https://stackoverflow.com/a/67765064/8548372) on macOS. \]

But yeah, this is really thorough! Thanks again! I would be okay if this merged as-is, but the comments are there. Very happy to be able to give this an approve.

src/auth.coffee
DeeDeeG: We have an opportunity to potentially clarify that this is the  "pulsar-edit.dev **Package** API Token", as opposed to the somewhat similarly named `atom-github` token for the `github` package. (Maybe we will rename this token to `pulsar-github` some day... ?).

So if anyone agrees with this suggestion, then feel free to update this tokenName to say "pulsar-edit.dev **Package** API Token". If no-one agrees then, LGTM as-is. Not a blocker IMO.

src/install.coffee
Daeraxa: Yep, looks like this was just missed.

README.md
```diff
@@ -1,23 +1,22 @@
-# apm - Atom Package Manager
+# ppm - Pulsar Package Manager
 
-![Build Status](https://github.com/atom-ide-community/apm/workflows/CI/badge.svg)
-[![Dependency Status](https://david-dm.org/atom/apm.svg)](https://david-dm.org/atom/apm)
+Discover and install Pulsar packages powered by [pulsar-edit.dev](https://web.pulsar-edit.dev).
```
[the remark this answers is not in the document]
Daeraxa: Yeah this has been my general strategy when dealing with the package site, I refer to the main website but link to the `web.pulsar-edit.dev` address so that it brings you to the actual package site, much in the same way that Atom would have the link displayed as `atom.io` going to `atom.io/packages`.

README.md
```diff
@@ -29,16 +28,16 @@ Therefore you can think of `apm` as a simple `npm` wrapper that builds on top of
...
-Run `apm help` to see all the supported commands and `apm help <command>` to
+Run `ppm help` to see all the supported commands and `ppm help <command>` to
 learn more about a specific command.
 
-The common commands are `apm install <package_name>` to install a new package,
-`apm featured` to see all the featured packages, and `apm publish` to publish
-a package to [atom.io](https://atom.io).
+The common commands are `ppm install <package_name>` to install a new package,
+`ppm featured` to see all the featured packages, and `ppm publish` to publish
+a package to [pulsar-edit.dev](https://web.pulsar-edit.dev).
```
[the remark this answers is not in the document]
Daeraxa: I personally think it is fine (same reason as the other comment) but if people feel otherwise then more than happy to change it.

src/auth.coffee
Daeraxa: I'll bow to your wisdom on this one to be honest so happy to make the change there (and it seems to have @confused-Techie's blessing too)

src/auth.coffee
Daeraxa: Again, happy to bow to wisdom here so if it is ok to make the change then lets do it.

DeeDeeG: Okay, the poll was up for a week on Discord, we got no-one voting to re-use old the Atom token, six people voting to use a new token, even if it meant they had to re-log-in once.

I already felt this was the right thing to do, but it eases the decision to get the poll results IMO.

DeeDeeG:
Okay, I'm gonna merge this! Thank you so much for the effort and responsiveness to the review comments, waiting for the poll, etc!

Reasons I think this is good to merge:

- Folks in the poll on Discord unanimously approved use of a new token.
- This has my approval (figuratively + I ticked the green check mark), @confused-Techie reacted positively in the review threads, so I see at least general agreement if not multiple formal approves.
- This has a couple super important things in it, especially moving away from `pulsar-edit.com` URLs (see #50).

Thanks again!

PR #56, opened 2023-02-13
fix: Don't assume `master` when checking git packages for upgrades

### LINKED ISSUE #55: `ppm upgrade` doesn't work for git-installed packages without a `master` branch
### Thanks in advance for your bug report!

- [X] Have you reproduced issue in [safe mode](https://pulsar-edit.dev/docs/atom-archive/faq/sections/what-does-safe-mode-do.html)?
- [X] Have you used the [debugging guide](https://pulsar-edit.dev/docs/atom-archive/hacking-atom/#debugging) to try to resolve the issue?
- [X] Have you checked our [FAQs](https://pulsar-edit.dev/docs/launch-manual/sections/faq/) to make sure your question isn't answered there?
- [X] Have you checked to make sure your issue does not already [exist](https://github.com/issues?q=is%3Aissue+user%3Apulsar-edit+is%3Aopen)?
- [X] Have you checked you are on the latest release of Pulsar?

### What happened?

I was wondering why some of my git-installed packages were able to be upgraded — both within the editor and via `ppm upgrade` — and some were not. I managed to track down what I think is the answer: `ppm upgrade` [checks a hard-coded `refs/heads/master`](https://github.com/pulsar-edit/ppm/blob/master/src/upgrade.coffee#L119), and many newer repos use `main` as the default branch.

GitHub changed the default on new repos back in 2020, but never updated `apm` accordingly.

I suppose the simplest fix would be to do `repo.getUpstreamBranch('refs/heads/main') || repo.getUpstreamBranch('refs/heads/master')`, so that’s probably what I’ll do unless someone has a better suggestion.

### Pulsar version

1.101.0-beta

### Which OS does this happen on?

🍎 macOS

### OS details

11.7.3

### Which CPU architecture are you running this on?

64-bit(x86_64)

### What steps are needed to reproduce this?

I don't expect anyone to do this, but here we go:

1. Create a GitHub repo with `main` as its default branch. Put a skeleton package in it — perhaps a sample package from the Flight Manual.
2. Install it with `ppm install [github username]/[repo name]`.
3. Land a trivial update to the package on `main`.
4. Attempt to upgrade it with `ppm upgrade [repo name]`; nothing will happen.

### Additional Information:

_No response_

savetheclocktower:
Fixes #55.

It took me ages to figure out how to replicate the `test-git-repo.git` fixture, but i managed to create one from scratch that uses `main` instead of `master`. All existing tests pass.


confused-Techie (approved):
The actual code changes here are simple and seem like a no brainer.

Literally three line changes from hardcoded values to dynamic ones that will continue to work in the future as far as we can reasonably guess.

But amazingly this PR also adds tests, new spec files, and everything needed to make sure this code works flawlessly. 

I'd honestly say this is a 10/10 PR, and fully say we should get it merged! Thanks a ton for the contribution!

DeeDeeG (approved):
LGTM!

I think there may be places elsewhere in ppm that expect `master` in a hard-coded way, but I think we all want to move away from that, and this is as good a start on that as any.

Thanks for the contribution!

savetheclocktower:
> I think there may be places elsewhere in ppm that expect master in a hard-coded way, but I think we all want to move away from that, and this is as good a start on that as any.

The good news there is that I didn't find any others. The only implicit assumption of `master` that remains is in the code that converts a TextMate bundle on GitHub to an Atom package, and I can't imagine anyone is writing new TextMate bundles these days that would have a `main` default branch.

PR #58, opened 2023-02-24
Fix installing with yarn on Windows

DeeDeeG:
### In short, this fixes `yarn install` for this repo on Windows

(Without breaking `npm install` for this repo, either.)

This PR fixes the scenario where you want to install this repo's dependencies with Yarn on Windows. (Should allow `yarn build:apm` to actually work on Windows over at the Pulsar core repo!) It does so by working around some differences with how _Yarn_ interacts with Windows' cmd.exe command line interpreter and batch scripts, compared to how _npm_ interacts with cmd.exe and batch scripts.

For technical explanation of each change, see the commit messages. Each change is no more than one line (and no more than 6 characters, at that!), other than whitespace. So do consider looking at the diff first, then referring to the commit messages if/only if you have further questions. 100% of the changes are found in two .cmd batch script files.

**That's the technical info, the rest of this PR body is commentary/not much more than a rant, but it gives context. Other than the footnote about npm major versions, which contains technical information I learned more about while preparing this PR, but which is a tangent and arguably off-topic. Feel free to skip reading the rest of this PR body if you like, since it shouldn't be technically relevant to reviewing the changes in this PR.**

### Background, effort, and why's cmd.exe gotta be this way?

This was surprisingly involved, as I wanted to know what I was doing with these fixes, but IMO cmd.exe's documentation is loosely worded, sparsely provided, scattered to the many corners of the internet, describes a generally janky/weird command execution environment / scripting language, and feels like a bunch of loosely cobbled together conventions accreted into a de-facto "don't touch it, it'll break someone's scripts" ball of stuff more than it feels like a (well or reasonably or even "at all") designed scripting and command execution environment.

CMD.exe's behavior feels more like happenstance, magic and luck (or misfortune, depending on your perspective) than a well-thought out scripting environment. IMO.

I'm not a fan.

### What did I do about it in this PR?

_Anyway_, this has not only a few tweaks to the .cmd scripts, but a _detailed explanation of the rationale of why I made these tweaks_. (See the commit message for each effectively one-line change for the detailed explanation).

I don't like changing small, obscure things because "eh, works for me on my machine", I want to do it because it's _the right solution_. Otherwise, the fix always feels like a temporary band-aid, and I spend all my time waiting for it to spring a leak or turn out to be woefully problematic on somebody's machine or with somebody's use-case. I want it done _once, and right_. Doing it the "works on my machine, don't think too hard about why" way always feels like how we got here, not like a way out from here.

_So_.

Here it is. This makes the postinstall scripts behave under `yarn install` and `npm install`\* alike.

\* Tested and working with npm 6.x only.

<details><summary><b>Compatibility note for npm major versions</b> (kind of a tangent/off-topic, click to expand if you want):</summary>

If/when we migrate to newer npm, we should do so once and never look back, since the major versions do not forward-and-backward interop well with each-other, though one major npm version range is mostly compatible with itself. Lockfile handling is different enough between the different major npm versions to be a major headache for this repo.

Particularly, the `jasmine-node` github URL gets saved in package-lock.json in one way with npm 8 or 9, that eventually breaks npm 6. npm 6 tries to reformat this github URL to suit its own way of doing things, but after it modifies and saves package-lock.json, npm 6 cannot handle what it wrote out, and it will error if you try to `npm install` with no `node_modules` folder populated (a fresh install).

Either we need to do some R&D on how to get that to stop happening, or interop between npm 6 and npm 8 or 9 will be a real nuisance, enough of a hassle to want to strongly avoid IMO. So we should stay on npm 6 for now, IMO, until we can make the jump to npm 8 or 9, and then not go back to npm 6.

npm 6.x is kind of sort of going end-of-life at the end of April 2023, when NodeJS 14 stops being supported, (https://github.com/nodejs/Release/#release-schedule), so we should look into this soon, IMO. (npm 6 is already sort of not supported, but it gets attention due to shipping with NodeJS LTS 14, support for which ends May 2023.) But upgrading npm is way out of scope for this PR, again.

</details>

### DIFF
--- a/script/postinstall.cmd
+++ b/script/postinstall.cmd
@@ -13,3 +13,5 @@ echo ^>^> Rebuilding apm dependencies with bundled Node !bundledVersion!
 setx JOBS 16
 
 call .\bin\npm.cmd rebuild
+
+exit /b
--- a/script/postinstall.cmd
+++ b/script/postinstall.cmd
@@ -3,7 +3,7 @@ setlocal EnableDelayedExpansion
 setlocal EnableExtensions
 
 echo ^>^> Downloading bundled Node
-node .\script\download-node.js
+node.exe .\script\download-node.js
 
 echo.
 for /f "delims=" %%i in ('.\bin\node.exe -p "process.version + ' ' + process.arch"') do set bundledVersion=%%i
--- a/bin/npm.cmd
+++ b/bin/npm.cmd
@@ -6,4 +6,4 @@ set "PATH=%~dp0;%PATH%"
 :: Force npm to use its builtin node-gyp
 set npm_config_node_gyp=
 
-"%~dp0\..\node_modules\.bin\npm.cmd" %*
+call "%~dp0\..\node_modules\.bin\npm.cmd" %*


DeeDeeG:
@confused-Techie I think you were wondering why `yarn install` in this repo or `yarn build:apm` in Pulsar core repo didn't work on Windows.

This PR fixes it, and my notes (commit messages) hopefully shed some light on what was going on if you want to know those sorts of answers as well.

So I figured I would tag you and let you know about this PR addressing the issue.


Best Regards,
\- DeeDeeG

confused-Techie (approved):
As always @DeeDeeG it's a delight reading your PR descriptions. And especially this time was surprised by the wonderful commit messages.

Seriously this deep dive looks like an intense one. Absolutely the definition of a programmer not being someone that types on a keyboard, but being someone that knows exactly what to type on a keyboard.

Impressive work, impressive researching. The final changes here all make sense and I have a hard time imagining any downside to these changes other than `ppm` finally working in a dev environment on Windows. 

As always it's very appreciated, I'll give this an approve and hope some others can review, otherwise I'd be much more than happy to see these changes merged

script/postinstall.cmd
```diff
@@ -13,3 +13,5 @@ echo ^>^> Rebuilding apm dependencies with bundled Node !bundledVersion!
 setx JOBS 16
 
 call .\bin\npm.cmd rebuild
+
+exit /b
```
DeeDeeG, line 17: So, this one honestly still confuses me the most.

<details><summary>Hidden behind an expander, because here lies madness (click to expand if you dare):</summary>

_(The other two I can sigh about and move on, it's just calling `.cmd` scripts from other `.cmd` scripts being super clunky and easily breaking things if not done just so. But at least the solution is at least kinda well documented... Some parts of it are a mystery to me, but only niche aspects of it, and I get the sense it's getting deep enough into it and there lies only madness and undefined/undocumented behavior beyond a certain point.)_

About `exit /b` ... the `/b` flag is only _supposed_ to mean "exit the current subroutine, not the whole `cmd.exe`." That much I can just about accept. (Why is this not the default though??? Why is exiting the entire `cmd.exe` the same lever as exiting a subroutine, and making the distinction is opt-in with the `/b` flag? Seems super error-prone... /grumpiness).

Here's the weird part, we're using it for something that I'm not sure it's supposed to do (hiding/resetting a previous error code).

It lets you explicitly exit with a certain error code, such as `exit /b 6` for error code 6, a number I just made up. _`exit /b` without an explicit code is supposed to carry forward the previous error code, if you don't explicitly set one like `exit /b [code]`._

Sources:
- https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/exit
  - Official docs, doesn't really clarify what happens to a previous/existing error code if you don't specify a new one.
  - (The official docs are way too short, IMO, they leave a lot of these finer questions un-addressed.)
- https://ss64.com/nt/exit.html
  - Some madman ([Simon Sheppard](https://ss64.com/docs/), apparently), who I have to trust implicitly, since manual testing seems to be the main way to verify _anything_ about `cmd.exe`, and this seems (???) to be the product of much manual testing and mad note-taking and scribbling.
    - > `EXIT` without an ExitCode acts the same as [`goto:eof`](https://ss64.com/nt/goto.html) and will not alter the [ERRORLEVEL](https://ss64.com/nt/errorlevel.html)
      - Also this (https://ss64.com/nt/goto.html), oh no:
      - > If the jump is successfully made <code><a href="https://ss64.com/nt/errorlevel.html">%ERRORLEVEL%</a> = <i>unchanged</i></code>, typically this will be 0 but if a previous command set an errorlevel, that will be preserved (this is a bug).
      - > **_(this is a bug)_** 🤦

And yet here I am doing `exit /b` here to suppress the error level from the `npm.cmd` --> `node_modules\.bin\npm.cmd` we are calling just above.

As far as I can read in any docs, this should do nothing. And yet I have tested it works _with_ this change and fails _without_ this change over and over. Due to the error level being visible to Yarn, and Yarn exiting non-zero, which would kill a CI run.

(Other than in CI, the error messages from the `npm rebuild` would just be a bit of a nuisance, since the C/C++ code seems to have built just fine??? I dunno how to verify all that built C/C++ code, honestly, but all tests are passing, right??? But failed CI runs is a bit of an issue, so I still think "hiding" the errors is better if they aren't actual problems. Like, maybe the `makefile` or equivalent for these C/C++ addons tries some commands that might be missing and does fallback behavior that works instead, but the original "command not found" error status persists long enough to trip up Yarn without this change??? I have to guess so hard at this, since it's increasingly deep down the rabbit hole just to fix a package manager issue.)

</details>

This, and having to scratch my head a lot to realize about the Yarn shims, were a large part of my descent into madness on this one (on this PR). I'm not even sure how/why npm avoids all these issues. And at this point, _I don't want to know_.

DeeDeeG (commented):
Annotating the PR diff with the particular changes I realized we needed.

Hopefully this is a lot easier to read and see what I was talking about, than the commit messages in the abstract.

script/postinstall.cmd
confused-Techie: Reading this really just makes me say, lets make all scripts JavaScript ones, with `#/env/node` at the top, and require everyone to have node already. That's just a lot of ridiculousness

DeeDeeG: I have loosely aspired to replacing these shell and batch scripts with pure JS for some time. I assume it would've been harder ~9 years ago when apm was first written, but NodeJS has kind of had everything and the kitchen sink built in by now. Indeed `child_process.spawn()` and the like can set env vars and call out to CLI commands, generally replacing the main point of shell/batch scripts.

If we can do JS with the same or better reliability for ppm in-use, then it'll be worth it. Otherwise, if the shell/batch scripts run more reliably, then perhaps the insanity these scripts bring during maintenance can be a maintainer-only concern, and a rare one (just don't touch them that often)... Not that I love the thought of that. We'll see...

_(More cmd weirdness: You might be intrigued to hear that adding `exit /b` at the bottom of our `bin\npm.cmd`, closer to where the errors I was trying to suppress came from, seemed to do nothing. Whereas doing `exit /b` one further layer out, at the bottom of `postinstall.cmd`, did the trick. 🤷.)_

Spiker985 (approved):

PR #60, opened 2023-03-08
meta: Actually sync yarn.lock

DeeDeeG:
This should make yarn.lock stable across multiple installs.

(Not sure why this diff wasn't part of the npm bump, but who knows?)

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -769,7 +769,7 @@ debug@^3.1.0:
   dependencies:
     ms "^2.1.1"
 
-debuglog@*, debuglog@^1.0.1:
+debuglog@^1.0.1:
   version "1.0.1"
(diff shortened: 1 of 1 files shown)


DeeDeeG:
Note:

This change is also committed in #57, but it is arguably a bit off-topic/out-of-scope there (?) and there is no diff between this branch and that one in terms of `yarn.lock`, so there should be no merge conflict if this one merges first.

#57 is currently a WIP PR with multiple changes needing to be done, and/or its scope reduced significantly to be mergeable, and so I don't think it'll be ready right away, whereas this is a really narrow one-file diff with a small diff at that.

If possible, I'd like to get the present PR merged here before syncing back to the core Pulsar repo again.

confused-Techie (approved):
CI looks great, diff looks good, lets get this one merged!

DeeDeeG:
Thank you for the review, merging this one, PR at core to bump the ppm submodule soon!

PR #62, opened 2023-03-24
Replace "apm" by "ppm" in help messages.

azuledu:
Replace all "apm" occurrences by "ppm" in console help messages.

### DIFF
--- a/src/auth.coffee
+++ b/src/auth.coffee
@@ -29,7 +29,7 @@ module.exports =
         else
           callback """
             No package API token in keychain
-            Run `apm login` or set the `ATOM_ACCESS_TOKEN` environment variable.
+            Run `ppm login` or set the `ATOM_ACCESS_TOKEN` environment variable.
           """
 
   # Save the given token to the keychain.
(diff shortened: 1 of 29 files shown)


Daeraxa:
Seems good to me.

One note semi-related to this PR but not to be actioned on this is that we really need to decide which out of `ppm` or `pulsar -p` is the "primary" way of interacting with ppm.
For example if you run some of these commands running `pulsar -p` it seems odd that it returns instructions to use `ppm` instead.

azuledu:
Maybe it is possible to substitute all "ppm" and "pulsar" occurrences in texts by a variable representing the script name. Something like `$0` in bash.

confused-Techie (approved):
LGTM! Thanks for doing this! And glad to see you contributing!

DeeDeeG:
This makes me want to finally rename the actual binaries in this repo from `apm` to `ppm`.

But that is just another thing that's not meant to be part of this PR itself.

DeeDeeG (approved):
Looks Good To Me™️.

If no objections, I intend to merge this PR soon.

confused-Techie (approved):
This looks fantastic!
Thanks a ton for contributing and getting this done! Sorry it wasn't reviewed sooner

confused-Techie:
@DeeDeeG Yeah that would be really awesome, and make things especially a little less weird on Windows as you can't use `pulsar -p`. So I'd be totally on board with that.

But with two approvals here, I'll merge this one

PR #19, opened 2022-11-06
Decaffeinate `spec/spec-helper` (Merge first: it's a dependency)

GuilleW:
Helper for unit test.
PR needed before other decaffeination in spec, because this file is included.
I don't remove `spec/spec-helper.coffee` actually, because it's useful during decaffeination.
⚠️ Should be removed at this end! ⚠️

PR #63, opened 2023-03-28
Decaffeinate spec helper (updated)

2colours:


2colours:
This is basically the updated version of https://github.com/pulsar-edit/ppm/pull/19, targeting the dedicated branch for decaffeinization. The only thing I changed on principle is the semicolons: I think we shouldn't go for the "automatic semicolon insertion" rules. Explicit is better than implicit with this surely.

If you think the coffee file shouldn't be removed, let me know and I will drop that commit.

confused-Techie:
@2colours This looks fantastic!

Totally agree, automatic semi-colon insertion I feel like is messy and harder to read, so a fan of them being put back in. 

I'm seeing you also got ride of the map to invoke `andCallThrough()`, not totally needed, but does make things a little less messy, and looks like it'd still do the exact same thing, so super rad!

And sure thing, removing the CoffeeScript at this point seems like a fine move to me, and with all that said 

* `spec/spec-helper.js`: ✅

So if you wanna continue with this process by all means, and hope you can see what I mean by how much easier it is to review.

confused-Techie (approved):
Now knowing the decaf PRs are being done separately I can review them individual. (Wasn't sure on the first review I left here).

But this PR looks good, all tests are passing, and the decaf work here isn't very extensive. If anything this PR is likely more a nice refactoring. But LGTM!

Thanks a ton for this contribution

PR #20, opened 2022-11-06
Decaffeinate `spec/apm-cli-spec`

GuilleW:
Merge https://github.com/pulsar-edit/ppm/pull/19 first.
EDIT : ❌ ~~[decaffeinate-project](https://decaffeinate-project.org) is used at first, then manual refactor/clean.~~

From coffee
```sh
$ jasmine-focused --captureExceptions --coffee spec/apm-cli-spec.coffee
> 6 tests, 24 assertions, 0 failures, 0 skipped
```
to js
```sh
$ jasmine-focused --captureExceptions spec/apm-cli-spec.js
> 6 tests, 24 assertions, 0 failures, 0 skipped
```

PR #64, opened 2023-03-28
Decaffeinate apm cli spec

2colours:


2colours:
The "successor" of https://github.com/pulsar-edit/ppm/pull/20 with a little twist: semicolons added and template strings preserved.

It's easier to run a diff in e.g VSCode, it maps the lines better.

confused-Techie (approved):
Light approval.

The code here looks solid. Some great decaf work opting for implicit functions rather than anonymous ones. Otherwise the other changes found don't seem of extreme significance, but mainly focusing on readability. 

Other than waiting for tests to complete, this should be good to do.

2colours:
> Light approval.
> 
> The code here looks solid. Some great decaf work opting for implicit functions rather than anonymous ones. Otherwise the other changes found don't seem of extreme significance, but mainly focusing on readability.
> 
> Other than waiting for tests to complete, this should be good to do.

Thank you for the feedback.

Honestly, I don't know the API very well, this `expect`, for example. If you could give me some pointers what further changes you'd like to see, I can dive deeper into it.

confused-Techie:
@2colours 
So for `expect` and `it` and things like `beforeEach`, `beforeAll` and that sort of thing you'll find in the `spec` files are all globals of `jasmine`.

Which is the testing framework listed here.

I'd love to link directly to some docs, but part of the hard thing is it looks like the Jasmine used in much of Atom's stuff was forked from a fork from a fork, ad nauseam. So means the docs aren't super clear.

But for understanding the basic principles, and the general syntax patterns we would expect feel free to look at [their docs](https://jasmine.github.io/)

confused-Techie (approved):
Alright, finally a proper review. This looks good, some simplistic changes, but approving, and will go ahead and merge.

PR #21, opened 2022-11-08
Decaffeinate `spec/ci-spec`

GuilleW:
Since I work on many files, I edit `package.json` to add useful shortcut to avoid misspelt filename.

```json
    "test-decaff": "npx coffee --compile --output spec spec/${npm_config_name}.coffee",
    "test-coffee": "jasmine-focused --captureExceptions --coffee spec/${npm_config_name}.coffee",
    "test-js": "jasmine-focused --captureExceptions spec/${npm_config_name}.js"
```

## Steps to reproduce decaffeination ([guidelines](https://github.com/pulsar-edit/.github/blob/main/project-birth/CONTRIBUTING-DURING-START.md#decaffeination)):

- [x] Check package.json for coffee-script version
  - version: 1.12.7
- [x] Decaffeination with official Coffee program
  - `npx coffee --compile --output spec spec/ci-spec.coffee`
  - (or my shortcut:  `./bin/npm run test-decaff --name=ci-spec`)
- [x] Quality of Decaffeination
  - [x] Replace `var` with `const` and `let` if it's possible.
  - [x] Replace Nested [Ternary Operators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator) with if statements.
  - [x] Remove `return` when it's not needed.
  - [x] Remove Result Arrays when they are not needed.
  - [x] Replace for loops with [for-of loops](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of).
  - [x] Use [Arrow Functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) to write functions in a single line when it's possible.
  - [x] Remove unneeded ref variables.
  - [x] Remove [do blocks](https://devdocs.io/coffeescript~1/index#loops) and use let or const instead.
  - [x] Tests and assertions should be the same and success like in `coffee` file
  
## Test from coffee and JS should output same successful result :

From coffee
```sh
$ jasmine-focused --captureExceptions --coffee spec/ci-spec.coffee
// or my shortcut $ `./bin/npm run test-coffee --name=ci-spec`
> 4 tests, 8 assertions, 0 failures, 0 skipped
```
to js
```sh
$ jasmine-focused --captureExceptions spec/ci-spec.js
// or my shortcut $ `./bin/npm run test-js --name=ci-spec`
> 4 tests, 8 assertions, 0 failures, 0 skipped
```

PR #65, opened 2023-03-28
Decaffeinate ci spec

2colours:


2colours:
Supersedes https://github.com/pulsar-edit/ppm/pull/21.

unused occurrances of `request` changed to `_request`, and one rather shouldn't `require` JSON files so reading it synchronously.

confused-Techie:
@2colours One question I have concerns about, since the decaf PRs will be merged to main, will it cause issues that they are all separate PRs? Since these changes would be totally fine to be all on one PR.

spec/ci-spec.js
```diff
@@ -12,133 +6,115 @@ const express = require('express');
 const wrench = require('wrench');
 const CSON = require('season');
 const apm = require('../lib/apm-cli');
-const {
-  nodeVersion
-} = require('./config.json');
+const { nodeVersion } = JSON.parse(fs.readFileSync('./config.json'));
```
confused-Techie, line 9:
```suggestion
const { nodeVersion } = require('./config.js');
```

Node is able to require JSON files without any extra effort on our part, so this one can stay as is.

spec/ci-spec.js
```diff
@@ -12,133 +6,115 @@ const express = require('express');
...
-    app.get('/test-module-with-dependencies', (request, response) => response.sendFile(path.join(__dirname, 'fixtures', 'install-locked-version.json')));
-    app.get('/test-module', (request, response) => response.sendFile(path.join(__dirname, 'fixtures', 'install-test-module.json')));
-    app.get('/native-module', (request, response) => response.sendFile(path.join(__dirname, 'fixtures', 'native-module.json')));
-    app.get('/tarball/test-module-with-dependencies-1.1.0.tgz', (request, response) => response.sendFile(path.join(__dirname, 'fixtures', 'test-module-with-dependencies-1.1.0.tgz')));
-    app.get('/tarball/test-module-1.1.0.tgz', (request, response) => response.sendFile(path.join(__dirname, 'fixtures', 'test-module-1.1.0.tgz')));
-    app.get('/tarball/native-module-1.0.0.tgz', (request, response) => response.sendFile(path.join(__dirname, 'fixtures', 'native-module-1.0.0.tgz')));
-
+    app.get(`/node/${nodeVersion}/node-${nodeVersion}-headers.tar.gz`, (_request, response) => response.sendFile(path.join(__dirname, 'fixtures', 'node-dist', `node-${nodeVersion}-headers.tar.gz`)));
```
confused-Techie, line 24: Is there a reason that we are using `_request` here rather than `request` like in the original decaf?

confused-Techie (changes requested):
Overall this PR looks good, there's one change that should be made, as it's generally the more performant option, and another change I wanted to address prior to merging.

Otherwise the tests are currently running and fingers crossed they look good, appreciate this a ton @2colours and sorry this last one took a bit to get to

spec/ci-spec.js
2colours: It indicates that the variable is unused. This is something that the default VSC linter respects, for example.

spec/ci-spec.js
2colours: Yes, it is able to but
1. it seems overall bad practice to mix up actual module importing (that might not even use CommonJS in the future, hopefully) and config loading
2. JSON undergoes stricter validation than any JS code

This was an intentional change, as indicated in the opening comment: `one rather shouldn't require JSON files so reading it synchronously`.

spec/ci-spec.js
```diff
@@ -12,133 +6,115 @@ const express = require('express');
...
 const CSON = require('season');
 const apm = require('../lib/apm-cli');
-const {
-  nodeVersion
-} = require('./config.json');
+const { nodeVersion } = JSON.parse(fs.readFileSync('./config.json'));
 
-describe('apm ci', function() {
+describe('apm ci', () => {
   let [atomHome, resourcePath, server] = Array.from([]);
```
2colours, line 12: On the other hand, this does look redundant to me in retrospect.

spec/ci-spec.js
confused-Techie: While I can see where you are coming from, you've also got to consider that the way finding the module/file is handled differently.

Currently all tests here are failing, because it can't find `./config.js`, in general `fs` will try to load a file from the path of where the script was called, in this case it's looking for `config.js` at the root of the repository, whereas using a `require` is able to find the file to load, relative to the path of the file requiring it.

So to get tests passing you'd either need to resolve the path to `config.js`, otherwise `require` it as it was done.

Up to you on the solution there, but just an fyi.

spec/ci-spec.js
confused-Techie: Fair enough, that sounds awesome to me. Just double checking

spec/ci-spec.js
2colours: First off, the file was named `config.json`, not `config.js` so I'm not sure how `config.js` came to be in the whole discussion.
Second, regarding the file resolution... I know that the path resolution is different but I couldn't find anything _importing this js file_, so I had to assume that it is invoked on its own. Now I read up on how Node deals with this, in the CommonJS world, one would probably use `__dirname` to be always accurate. Unfortunately, there doesn't seem to be a catch-all way to do this; it either works for CommonJS or ESM. For the time being, perhaps CommonJS will do. I still prefer the "read the file and parse it" approach over hiding behind a `require` that ties our hands much more to the old way of things.

confused-Techie: Yeah sorry I misspoke there (This is why I shouldn't code review at 3am lol).
But I do get not wanting to hide this logic and more strictly parse it with an actual JSON parser. You may potentially be able to inspect the `process` values to see if the current file directory is exposed. But yeah we could stick with the `require` for now to get it working, and ideally later on find a way to do this better

2colours: I added a commit with the usual "append the filename to the __dirname variable" logic, already present in this script. Please check it out.

confused-Techie (approved):
With the important tests passing, and the changes here looking good I'll merge this one to the decaf branch.

PR #66, opened 2023-04-08
Decaffeinate clean spec

2colours:


2colours:
This is meant to supersede https://github.com/pulsar-edit/ppm/pull/22.
Like with https://github.com/pulsar-edit/ppm/pull/65, I was trying to stay away from `require`-ing the config file while incorporating the configurable node version commits. Also, eliminated some closure variables with their initialization.

confused-Techie (approved):
The changes here look good, and the relevant tests are passing

DeeDeeG (approved):
👍 

As @confused-Techie said, the tests pass and the changes look okay.

I don't totally understand the aversion to requiring a JSON file, but I suppose the change isn't worth worrying about either. EDIT: It would be too much hassle to start worrying about this in this PR when the approach was already accepted in #65, so disregard this comment, I guess.

There is a redundant commit deleting spec/spec-helper.coffee even though it is also deleted on the target `decaf` branch, this will add a little noise to the commit log but beyond that it is a non-issue, just something I saw and will point out as I'm reviewing. Again, I don't think this needs to change -- I would be fine merging this as-is. If in the future you rebase commits for any decaf PRs, you can leave out the commit deleting `spec/spec-helper.coffee` for cleanest/simplest commit history. But not a huge deal.

spec/clean-spec.js
```diff
@@ -1,98 +1,74 @@
...
+  it('uninstalls any packages not referenced in the package.json', () => {
     const removedPath = path.join(moduleDirectory, 'node_modules', 'will-be-removed');
     fs.makeTreeSync(removedPath);
-    fs.writeFileSync(
-      path.join(removedPath, 'package.json'),
-      '{"name": "will-be-removed", "version": "1.0.0", "dependencies": {}}',
-      'utf8'
-    );
-
+    fs.writeFileSync(path.join(removedPath, 'package.json'), '{"name": "will-be-removed", "version": "1.0.0", "dependencies": {}}', 'utf8');
```
DeeDeeG, line 52: Maybe 5 lines is a lot for something as simple as writing out some JSON to a file on disk. But I do think the indented and spaced out version was easier to read before being condensed onto one line.

I would not block this PR from merging for this, but if it were me I would leave the indentation for readability (easier to skim quickly IMO).

If you do simply undo this whitespace-only change, I would re-approve as soon as I see the notification (which may not be right away) but other reviewers can presume I am 👍 approving of said change.

spec/clean-spec.js
```diff
@@ -1,98 +1,74 @@
...
+  it('uninstalls a scoped package', () => {
     const removedPath = path.join(moduleDirectory, 'node_modules/@types/atom');
     fs.makeTreeSync(removedPath);
-    fs.writeFileSync(
-      path.join(removedPath, 'package.json'),
-      '{"name": "@types/atom", "version": "1.0.0", "dependencies": {}}',
-      'utf8'
-    );
-
+    fs.writeFileSync(path.join(removedPath, 'package.json'), '{"name": "@types/atom", "version": "1.0.0", "dependencies": {}}', 'utf8');
```
DeeDeeG, line 65: If undoing the above indentation change, please do this one as well. (If leaving them, that is also okay.)

2colours:
> I don't totally understand the aversion to requiring a JSON file, but I suppose the change isn't worth worrying about either. EDIT: It would be too much hassle to start worrying about this in this PR when the approach was already accepted in #65, so disregard this comment, I guess.

Anyway... I would hope that I said my biggest arguments under that PR - of course, these are still rather minor things but then the change itself is rather minor, too. It's a bit bothering that the code needs to rely on CommonJS either way. :(



> There is a redundant commit deleting spec/spec-helper.coffee even though it is also deleted on the target `decaf` branch, this will add a little noise to the commit log but beyond that it is a non-issue, just something I saw and will point out as I'm reviewing. Again, I don't think this needs to change -- I would be fine merging this as-is. If in the future you rebase commits for any decaf PRs, you can leave out the commit deleting `spec/spec-helper.coffee` for cleanest/simplest commit history. But not a huge deal.

Yep... honestly, I never quite understood merge. I understand rebase and I thought the outcome would essentially be the same but apparently not, yuck. I just wanted to avoid PRs building on top of each other.
I don't know if you can rebase these branches but even at the end, I suppose an interactive rebase can help clear up unwanted commits, in case.

spec/clean-spec.js
2colours: Okay, fair enough I think, I have undone this change.

DeeDeeG (approved):
Same approval reasons as before, plus requested whitespace-only change was made. Thank you. 👍

PR #25, opened 2022-11-08
Decaffeinate `spec/develop-spec`

GuilleW:
Since I work on many files, I edit `package.json` to add useful shortcut to avoid misspelt filename.

```json
    "test-decaff": "npx coffee --compile --output spec spec/${npm_config_name}.coffee",
    "test-coffee": "jasmine-focused --captureExceptions --coffee spec/${npm_config_name}.coffee",
    "test-js": "jasmine-focused --captureExceptions spec/${npm_config_name}.js"
```

## Steps to reproduce decaffeination ([guidelines](https://github.com/pulsar-edit/.github/blob/main/project-birth/CONTRIBUTING-DURING-START.md#decaffeination)):

- [x] Check package.json for coffee-script version
  - version: 1.12.7
- [x] Decaffeination with official Coffee program
  - `npx coffee --compile --output spec spec/develop-spec.coffee`
  - (or my shortcut:  `./bin/npm run test-decaff --name=develop-spec`)
- [x] Quality of Decaffeination
  - [x] Replace `var` with `const` and `let` if it's possible.
  - [x] Replace Nested [Ternary Operators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator) with if statements.
  - [x] Remove `return` when it's not needed.
  - [x] Remove Result Arrays when they are not needed.
  - [x] Replace for loops with [for-of loops](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of).
  - [x] Use [Arrow Functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) to write functions in a single line when it's possible.
  - [x] Remove unneeded ref variables.
  - [x] Remove [do blocks](https://devdocs.io/coffeescript~1/index#loops) and use let or const instead.
  - [x] Tests and assertions should be the same and success like in `coffee` file
  
## Test from coffee and JS should output same successful result :

From coffee
```sh
$ jasmine-focused --captureExceptions --coffee spec/develop-spec.coffee
// or my shortcut $ `./bin/npm run test-coffee --name=develop-spec`
> 3 tests, 12 assertions, 0 failures, 0 skipped
```
to js
```sh
$ jasmine-focused --captureExceptions spec/develop-spec.js
// or my shortcut $ `./bin/npm run test-js --name=develop-spec`
> 3 tests, 12 assertions, 0 failures, 0 skipped
```

PR #69, opened 2023-04-08
Decaffeinate develop spec

2colours:
Updated version of https://github.com/pulsar-edit/ppm/pull/25. No functional changes (this script hasn't been updated since the original PR) , semicolons readded.


spec/develop-spec.js
```diff
@@ -1,84 +1,75 @@
...
-      expect(fs.existsSync(linkedRepoPath)).toBeTruthy();
-      return expect(fs.realpathSync(linkedRepoPath)).toBe(fs.realpathSync(repoPath));
+  describe("when the repository hasn't been cloned", () => {
+    it('clones the repository to ATOM_REPOS_HOME and links it to ATOM_HOME/dev/packages', () => {
+      const Develop = require('../lib/develop');
+      spyOn(Develop.prototype, 'getRepositoryUrl').andCallFake((packageName, callback) => {
+        const repoUrl = path.join(__dirname, 'fixtures', 'repo.git');
+        callback(null, repoUrl);
+      });
+      spyOn(Develop.prototype, 'installDependencies').andCallFake(function (packageDirectory, options) {
```
confused-Techie, line 44:
```suggestion
      spyOn(Develop.prototype, 'installDependencies').andCallFake((packageDirectory, options) => {
```

confused-Techie (changes requested):
Tests are passing here, and overall the changes look good. Just pointed out a trivial, non-required change, but would likely be best to follow modern practices. But thanks for the contributions here

spec/develop-spec.js
2colours: Uh oh, this is a breaking suggestion. `this` is dynamically bound in `function` literals but lexically bound in arrow literals so the latter is almost certainly faulty; there is no `this`˙introduced anywhere lexically.
It needs to stay with `function` to have any hope of working properly.

confused-Techie: Seems interesting it's causing a failure, as I haven't run into it before.

But since that was only a stylistic suggestion before, in that case I'll close the suggestion. Thanks for pursuing it!

confused-Techie (approved):
LGTM!

spec/develop-spec.js
2colours: Well, maybe you also learned something, then. :P https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#cannot_be_used_as_methods

PR #70, opened 2023-05-20
Decaffeinate disable spec

2colours:
Meant to supersede https://github.com/pulsar-edit/ppm/pull/26.
Mostly taken verbatim since no modifications happened to the affected file since. I readded the commas as usual.


spec/disable-spec.js
```diff
@@ -1,128 +1,106 @@
...
+    CSON.writeFileSync(configFilePath, {
+      '*': {
+        core: {
+          disabledPackages: ['test-module']
+        }
       }
-    }
-  });
-
+    });
```
DeeDeeG, line 25: I know this has been done in the previous PRs, but as a personal preference I don't like stripping out all the blank lines. I feel that breaking up separate ideas in the code with blank lines helps with readability when reading the code over for the first time, which IMO is the most important read of any code you will do, reading it over the first time around.

I wonder if @confused-Techie has any thoughts on this? If they don't outright feel the same way about it then I might just leave this as-is for consistency with the previous merged PRs.

At least knowing that we committed the machine decaf first, we can go back and see what exactly changed compared to the machine decaf. If I feel really strongly about it in the future I can go back and re-add the blank lines.

So as I said if @confused-Techie or another reviewer agrees with leaving the blank lines then we can standardize on that going forward, otherwise don't worry about it.

DeeDeeG (approved):
Changes look good. As in, no "meaningful" changes, which is what I would be hoping for out of a decaf-related PR. Just coding style and cleanup and things like that.

I don't see any failing specs in the CI output, either.

confused-Techie (approved):
LGTM! Passing tests clean diff, appreciate the efforts!

PR #75, opened 2023-05-21
Decaffeinate init spec

2colours:
Meant to supersede https://github.com/pulsar-edit/ppm/pull/31.
Again, no upstream changes; semicolons and `let` brought back.


confused-Techie (approved):
LGTM! Gotta say, thanks for getting rid of the multi-line template literals, for some reason to me, those always make things harder to read than easier, so I'm on board with that change personally.

But otherwise changes look good and tests are passing

PR #76, opened 2023-05-21
Decaffeinate install spec

2colours:
This pull request is meant to supersede https://github.com/pulsar-edit/ppm/pull/32 but it wasn't that much a mechanic task.
There have been upstream changes, and the complexity of the code did show: the former caffeination showed some archaic structures (like the `arguments` magic argument that needs to be upgraded to an Array). Also, quite a lot of comments were missing in it so what I did was more or less a review in itself, keeping good changes and reverting obsolete ones while making sure nothing gets lost.


confused-Techie (approved):
LGTM

PR #79, opened 2023-07-12
Switch to our npm fork, to get newer node-gyp (node-gyp 9.x)

DeeDeeG:
Move to our fork of npm, which changes nothing other than depending on newer node-gyp. For ppm, this effectively bumps node-gyp from 5.1.1 to ^9.4.0.

This adds support for newer OSes/compiler toolchains/etc. (And _drops_ support for Python 2.x, and Python 3.6 or older.)

See the node-gyp changelog for more info:
https://github.com/nodejs/node-gyp/blob/main/CHANGELOG.md

This node-gyp upgrade has been a wishlist item for the development team (and others?) for quite a while now.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -34,7 +34,7 @@ jobs:
           check-latest: true
 
       - name: Install dependencies
-        run: npm install
+        run: yarn install
 
       - if: "!contains(matrix.os, 'windows')"
         name: Run tests 👩🏾‍💻
--- a/package-lock.json
+++ b/package-lock.json
@@ -33,6 +33,11 @@
           "integrity": "sha512-n5M855fKb2SsfMIiFFoVrABHJC8QtHwVx+mHWP3QcEqBHYienj5dHSgjbxtC0WEZXYt4wcD6zrQElDPhFuZgfA==",
           "dev": true
         },
+        "is-fullwidth-code-point": {
+          "version": "3.0.0",
+          "resolved": "https://registry.npmjs.org/is-fullwidth-code-point/-/is-fullwidth-code-point-3.0.0.tgz",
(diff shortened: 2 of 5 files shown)


confused-Techie:
@DeeDeeG mind me adding a commit to your branch here that switches our CI to use `yarn` as we discussed on Discord?

DeeDeeG:
@confused-Techie I really wanna get this one working with npm also, but... Maybe I should do that as follow-up work when I get around to it. Perhaps it shouldn't hold this all up.

Like a silly little PR to "fix package-lock.json that we totally don't use so npm 8+ users who we officially totally do not support don't get bothered by this being broken for them"

DeeDeeG:
> @DeeDeeG mind me adding a commit to your branch here that switches our CI to use `yarn` as we discussed on Discord?

Ok 👍

confused-Techie:
@DeeDeeG If that's something your comfortable with, that sounds awesome to me. Since once we get this in, I was hoping to myself get around to some of the changes I've been meaning to make in this repo for a while, such as decaffing and what not. So if we get everything merged I can do that real fast

confused-Techie:
@DeeDeeG So interestingly enough, we are getting consistent failures on Windows NodeJS 16 only, it's failing to build oniguruma

confused-Techie:
Alright, so over on #82 I've tested some solutions here.

It seems that if we migrate to `second-mate` and get an updated oniguruma Windows (along with all other platforms) are able to build everything just fine with these changes. Except, since `await` doesn't seem to be supported in CoffeeScript 1.x (which we are using) we can't actually migrate to `second-mate` as that requires proper async support.

Sure we could figure out how to get that code implemented, but it may be better to just let this one spec fail for now and merge, then once we have decaffed our code, we can migrate to `second-mate` with proper async support, and then we can get everything happy here.

So that is to say, I think we should just merge with our one failure here

confused-Techie (approved):
Gonna merge this one now, so we can decaf with these changes, and migrate to second mate to get all versions of windows working properly, like mentioned above

PR #84, opened 2023-07-23
Decaf Source

confused-Techie:
This PR decafs the entire source of PPM. Following with our current guidelines of initially comparing against a machine decaf branch. Once ready to merge, the target branch will be changed to `master`. This is done to make review possible, as there's something to compare changes against.

The vast majority of this decaf is rather light, this is done on purpose, since I'd rather not break anything, especially considering that the vast majority of modified values are passed around rather than originating in any particular function.


confused-Techie:
So knowing the tendency of decaf PRs to properly have a review, I may just go ahead and trust the tests on this one, since there are a few other changes I'd actually like to make beyond just decaf work.

So knowing that all tests are passing here except "Windows - NodeJS 16" and "Windows - NodeJS 18" (Which are the fault of windows being unable to build the native modules we use after our upgrade to node-gyp, which should be solved once we can switch over to `second-mate` from `first-mate` since we had to do the same to bump NodeJS versions over in the Pulsar repo).

But this means the failures there are excepted, and otherwise all tests are passing totally fine.

So I'll go ahead and change the target branch to `master` and let our tests run to ensure everythings fine, then merge. Thanks to @DeeDeeG for all the assistance in getting things sorted over here.

PR #85, opened 2023-07-23
Cleanup `visualStudioIsInstalled()`

confused-Techie:
While decaffing the code, I found a very ugly unmaintainable piece of code, that I couldn't just let it stay there. 

So this PR rewrites it to be much cleaner, much more understandable, easier to maintainer, and easier to add new values to, if needed.

Should result in zero logic changes.


confused-Techie:
I'd like to mention, the single failure here is not the fault of this PR, and is resolved over on #86

PR #86, opened 2023-07-23
Migrate to `second-mate` and remove `first-mate`

confused-Techie:
This PR migrates to `second-mate` and removing `first-mate`.

We had to do the same over on the Pulsar repo, and from earlier testing, it seems this should fix the build issues on both Windows platforms, that are currently failing after bumping `node-gyp` on our fork of NPM.

But even beyond these changes, we are no longer maintaining `first-mate` so if we want to have any updates made to the `-mate` library, we need to switch anyway.


savetheclocktower:
This is for the best. If we ever rewrite `ppm` from the ground up, we can almost certainly lose the `second-mate` dependency altogether, since its only purpose here is to convert TextMate themes and packages, and that's not exactly an important feature these days.

src/package-converter.js
```diff
@@ -199,6 +199,9 @@ class PackageConverter {
       }
 
       try {
+        (async () => {
+          await ready;
+        })();
         if (scope) { selector = new ScopeSelector(scope).toCssSelector(); }
       } catch (e) {
```
DeeDeeG, line 206: Maybe this is a noob JS question, and pardon for commenting on such an old PR, but does this really `await` the `ready` function and not run the `if` block until after the `await` finishes? Or is this some kind race condition?

Shouldn't the `if (scope) { . . . stuff . . . }` be _inside_ the immediately-invoked-(async)-function-expression thingy's (["IIFE"](https://developer.mozilla.org/en-US/docs/Glossary/IIFE)'s) block, on the line just under `await ready`? 

Or am I overthinking it, and the `if (scope) { . . . stuff . . . }` will be implicitly wrapped in a `Promise.then()`?

I feel like looking at the main Pulsar PR (https://github.com/pulsar-edit/pulsar/pull/435/files#diff-3174f13c059533f7db5695822c687ae324c3023722e249e0aad32afa18746554), the `await` was _just above_ the following code and looked more correct-ish to my (possibly noob) eyes.

DeeDeeG (commented):
Not sure about some of this syntax?

(See specific code highlighted comment, and two more places in the code with similar code pattern.)

src/package-converter.js
confused-Techie: So on this, from what should be happening here:

The `ready` and the actually library being usable aren't necessarily needed with each other, we simply have to ensure that the async `ready` has returned before we can use the library, so being right above is valid, but this should be as well.

Although, moving this bit into the self calling function would also be totally valid. From my understanding using a self calling function like this is essentially sync, letting this function return before continuing. Then once it returns, we move on with our other code. I only kept them separate like this, since the `selector` being defined itself isn't async. So it shouldn't matter to much.

Although this is absolutely one of the things I'd like to clean up with work like #88 but of course that'll take some time. But from my experience this should still work, and luckily tests are happy

DeeDeeG: > its only purpose here is to convert TextMate themes and packages, and that's not exactly an important feature these days.

I hope this TextMate stuff is exercised in the specs? Some former member of Atom team was wary of how complete the tests were for this repo once when I was talking with them, so I don't take full test coverage of all fetures for granted at ppm repo.

DeeDeeG: > `spec/init-spec.js:    describe('when converting a TextMate bundle', () => {`

Okay, yeah, I think TextMate stuff has specs... Thank goodness.

confused-Techie: Well would you look at that @DeeDeeG.
You are 100% right. There are absolutely 0 tests for this feature. So passing tests isn't the golden ticket I thought it was.

Oof. Lemme see if I can try these changes on the master branch

savetheclocktower: This block of code doesn't seem like it does anything. It's an `async` function that awaits and then does nothing. I don't think it'll have the intended effect.

confused-Techie: @savetheclocktower It was a concern to see there's no tests here.

But prior to bumping the version of PPM used in core, I manually ran the code to build a grammar file from a TMBundle, and things seemed to work as we would expect there. Maybe it's a red herring, but fingers crossed this is working as I hoped it would.

Since we don't need anything to happen in the async function, we just need to wait until `ready` then we can create the `ScopeSelector` otherwise, it can't be created prior to `ready`

savetheclocktower: This code won't wait to create the `ScopeSelector`. It schedules a promise and then immediately assigns the selector.

confused-Techie: > This code won't wait to create the `ScopeSelector`. It schedules a promise and then immediately assigns the selector.

Oof, well I may misunderstand how this behaves then. Thought generating the package locally would be a way to confirm. Might have to re-evaluate this as soon as I can to see exactly what goes on. Ideally by actually writing some tests that look at this file, which I'm still surprised to find out doesn't yet exist

confused-Techie: Otherwise, as this is an underutilized feature, I am still hoping to turn this repo into async, and at the very least it's comforting to know that doing this will fix the above issue

savetheclocktower: It's also not at all important to keep this feature. I cannot imagine that anyone has used it in years. I'm not saying we should proactively rip it out, but if we're spending a lot of time on retaining the TextMate-theme–conversion functionality, then we've gone astray.

confused-Techie: > It's also not at all important to keep this feature. I cannot imagine that anyone has used it in years. I'm not saying we should proactively rip it out, but if we're spending a lot of time on retaining the TextMate-theme–conversion functionality, then we've gone astray.

I do agree here to a point. While there should be some effort to ensure it still works, I totally think you are right in that it's been years since it was actually used. So I'll try to spend a negligible amount of time on building some way to test this. Maybe as simple as just running the command in a test to build a package, against a non-good creation of the same package, then compare the created files. Might be a dirt cheap and non-time intensive way to help ensure things do always properly work, without spending a lot of effort on doing so

PR #88, opened 2023-07-24
Convert to Async

confused-Techie:
IMO looking at this code does feel a lot like callback hell, so I'm hoping to remedy this by implementing async capabilities into PPM. Obviously we may not be able to do it all at once, so will start small and work up to the bigger portions of the code as possible.


confused-Techie:
As this PR has only failing tests, and PR #95 is seeing much more activity and success, I'll go ahead and close this one out in favour of #95

PR #90, opened 2023-09-05
Update unpublishing wording

Daeraxa:
Conversation thread here: https://discord.com/channels/992103415163396136/1147891572533362779/1147891572533362779

We decided it would be best to clarify the wording as it isn't clear from the text in ppm that the unpublish command will make the package name permanently unavailable.

This just adds a line to clarify that.


confused-Techie (approved):
Assuming all tests are happy, looks great to me! Thanks for following up on this one!

PR #91, opened 2023-09-08
Skip update checks for bundled packages

confused-Techie:
This PR allows PPM to skip checking for updates for bundled packages within Pulsar.

This only works if the package is not a valid git repository, which is true for the vast majority of our bundled packages.

Although, some outliers to this exist, those can be determined by checking the remaining items [here](https://github.com/pulsar-edit/pulsar/issues/512). As for all other of the 81 bundled packages, this should allow them to be silently skipped. Assumed to always be up to date by PPM.

This is done by checking the `repository` of the `package.json` of the package and determining if it's the same as the repository of Pulsar itself.

### DIFF
--- a/spec/upgrade-spec.js
+++ b/spec/upgrade-spec.js
@@ -76,6 +76,23 @@ describe('apm upgrade', () => {
     });
   });
 
+  it('does not display updates for "core" packages', () => {
+    fs.writeFileSync(path.join(packagesDir, 'core-package', 'package.json'), JSON.stringify({
+      name: 'core-package',
+      version: '1.0',
+      repository: 'https://github.com/pulsar-edit/pulsar'
+    }));
+    const callback = jasmine.createSpy('callback');
+    apm.run(['upgrade', '--list', '--no-color'], callback);
+
+    waitsFor('waiting for upgrade to complete', 600000, () => callback.callCount > 0);
+
+    runs(() => {
+      expect(console.log).toHaveBeenCalled();
+      expect(console.log.argsForCall[1][0]).toContain('empty');
+    });
+  });
+
   it('does not display updates for packages whose engine does not satisfy the installed Atom version', () => {
     fs.writeFileSync(path.join(packagesDir, 'test-module', 'package.json'), JSON.stringify({
       name: 'test-module',
--- a/src/upgrade.js
+++ b/src/upgrade.js
@@ -93,6 +93,12 @@ available updates.\
     }
 
     getLatestVersion(pack, callback) {
+      // We want to bail on checking for updates of any packages that come with the editor
+      // This can generally be detected by checking if the repository is that of Pulsar itself
+      if (pack.repository === "https://github.com/pulsar-edit/pulsar") {
+        return callback();
+      }
+      
       const requestSettings = {
         url: `${config.getAtomPackagesUrl()}/${pack.name}`,
         json: true


DeeDeeG (approved):
As a reviewer for _this repo_:

I'm sort of surprised none of the tests complain about this, although I guess exercising this specific scenario is outside the scope of any tests at the moment.

Ideally we might want to add a test for this, maybe as a follow-up perhaps, but I'm not feeling like asking for that for an approve on this, personally. If you _want to_, that'd be cool.

As a reviewer for _pulsar-edit org in general_:

Ideally we revert this as soon as we find a way to not ask ppm for these updates for core packages properly?? Since that seems like a cleaner location for the change to go, over in the requesting-the-check side, not the implementing-the-check side.

For now, I think this is worth doing, and this is a small change scoped specifically to our in-house repos. (We don't generate that `repository` URL in new package templates, right??)

Alrighty, so with that said, approved. 👍

DeeDeeG:
If you wanna address some of the feedback, then I can wait, otherwise, I am inclined to merge.

And once I review whatever all else has been committed since `ppm` was synced to core repo, do a `ppm` bump in core repo.

confused-Techie:
@DeeDeeG So on what you've mentioned here.

My other PR over in `pulsar` doesn't address what you mentioned.

The difficult part about just not asking ppm to check for updates of certain packages, is that `settings-view` actually doesn't ask for updates to any specific packages.

Instead `settings-view` just runs the generic command to ask PPM to check for updates for all installed packages. Which would include our bundled ones. So really this may be the only way to stop this behavior from occurring.

As for adding a test, I'd be more than happy to take a look to see if we can do that. But I'd like to say, I don't see any way we can easily revert this PR later down the road without significant changes.

DeeDeeG (approved):
CI is green with the new test, which is a little above and beyond for this kind of quick fix, even if I did request it. Thank you.

And thank you for the context replying to my previous review comments.

Tests pass (including new test), LGTM! Re-approving.

confused-Techie:
@DeeDeeG Exactly, which is unfortunate. But yeah `settings-view` doesn't even know what packages are being checked. So it'd be a significant code change to even make it aware of the packages to them go ahead and not check them. But hopefully this fix isn't too hacky to remain in the codebase for a bit of time.

If we really worried about it's implementation, we could add a flag to the command, such as `--skip-core` which would optionally enable this feature, so that it worked here, but otherwise wouldn't provide any strange behavior for end users

confused-Techie:
Lets merge, woot!

DeeDeeG:
I dunno if we want to revert the change, considering this is such a small and not-hurting-anything change, but it also isn't doing the thing it's meant to be doing in real-world testing, I think. (Given some further research into things since this was merged.)

confused-Techie:
@DeeDeeG Great catch here, and now that I think about it, you are totally right. This wouldn't ever see any packages that are bundled.

So yeah this is kinda pointless, and while harmless, I'd hate for it to add uncertainty into how PPM behaves to someone reading it in the future. So seems best to revert this PR.

DeeDeeG:
There's a shiny "Revert" button net to the "merge" event if scrolling up the history of this PR.

I'm somewhat curious to try it. Hmmmmmmmmmmmm. I might try that in a bit if no objections, or if you want to try it, go for it.

confused-Techie:
@DeeDeeG I've never seen this option. So that sounds like a plan then, lets give it a wirl

PR #92, opened 2023-09-11
Remove remnants of Coffeescript building

2colours:
Since there is nothing remaining for Coffescript, the linter file is not required and the readme shouldn't mention the build step that doesn't even exist anymore.


confused-Techie (approved):
Thanks a ton for contributing!

This one looks great to me, and has happy tests

DeeDeeG (approved):
I think we can also remove `coffee-script` and `coffeelint` from `packge.json`. They were listed as `devDependencies`, so not guaranteed to be present "in production".

I tried removing them, and tests still pass locally.

(So, even though we can generate packages that are CoffeeScript, it seems there's nothing about that that ppm needs to actually understand or convert CoffeeScript to be able to do. Just copying some files at that point, I suppose. I think Pulsar core handles actually reading those packages' CoffeeScript code.)

Would be open to a second opinion, but seems like as good an opportunity as any to get rid of those. And it doesn't have to be in this PR, we could do it as a follow-up. Just something I thought of when reviewing this PR.

Other than that, Looks Good To Me!

confused-Techie:
@DeeDeeG didn't realize those were still present in our `devDependencies` but yeah lets tackle that in another PR and get this one in for now. But I don't see any reason to worry about removing those

confused-Techie:
But with two approvals lets go ahead and merge this one

PR #96, opened 2023-09-17
Revert "Skip update checks for bundled packages"

confused-Techie:
Reverts pulsar-edit/ppm#91

We have discovered later on that this change doesn't actually need to happen.

Basically while the code is solid, the location packages are checked from to then check them for updates, would never include any actual bundled packages. Since those are all wrapped up in an asar, rather than the users `.packages` directory. 

So lets go ahead and revert it, to avoid complexity for complexities sake.

### DIFF
--- a/spec/upgrade-spec.js
+++ b/spec/upgrade-spec.js
@@ -76,23 +76,6 @@ describe('apm upgrade', () => {
     });
   });
 
-  it('does not display updates for "core" packages', () => {
-    fs.writeFileSync(path.join(packagesDir, 'core-package', 'package.json'), JSON.stringify({
-      name: 'core-package',
-      version: '1.0',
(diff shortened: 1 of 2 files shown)


DeeDeeG (approved):
Approve, reasoning was laid out more fully in a comment on the original PR: https://github.com/pulsar-edit/ppm/pull/91#issuecomment-1721462378

Thanks for getting to this! And for the overall efforts to address unneeded backend API usage in Pulsar / ppm. 👍 

Aaaaand... It's neat that the "Revert" button works! Could be convenient in the future, possibly maybe. _Who knows??_

DeeDeeG:
Approved, tests pass, this is a pretty simple revert PR, we can merge it IMO. Merging! Thanks again.

PR #99, opened 2023-09-24
Fix Error Handling

confused-Techie:
It was brought to my attention on [Discord](https://discord.com/channels/992103415163396136/992110062992621598/1155591813785194626) that a61aa7e121bc93723f675198778b6006fdd66738 broke error handling in PPM.

The clue to find out what had gone wrong was the fact that the unstar command still worked.

This is because in that PR I had told `superagent` a list of status codes to not error on. This means that when we got, say a 404, the actual `error` object returned to the callback would be `null` even though those commands then relied on the `error` object to determine what had gone wrong. 

So what I've done, is provide both the `error` and `body` to the `request.getErrorMessage()` so that we can check for errors in either of them. 

Meaning when a request fails we will:
1. Look first for within the `superagent` error object for `err.response.body` then for `err.response.error` then for `err`
2. If that fails, we can then look at the body for `body.message` then `body.error` then finally `body`

This helpfully means we prioritize any errors returned by `superagent` then any errors returned by the backend. And if all else fails, we just tell the user whatever message was returned to us by the backend, which can be helpful in case the backend's error schema ever changes by accident or on purpose and isn't updated here. So that way we can always inform the user exactly what the backend is thinking.

### DIFF
--- a/src/develop.js
+++ b/src/develop.js
@@ -61,7 +61,7 @@ cmd-shift-o to run the package out of the newly cloned repository.\
             return callback(`No repository URL found for package: ${packageName}`);
           }
         } else {
-          const message = request.getErrorMessage(error);
+          const message = request.getErrorMessage(body, error);
           return callback(`Request for package information failed: ${message}`);
         }
       });
--- a/src/featured.js
+++ b/src/featured.js
@@ -47,7 +47,7 @@ List the Pulsar packages and themes that are currently featured.\
           packages = _.sortBy(packages, 'name');
           return callback(null, packages);
         } else {
-          const message = request.getErrorMessage(error);
+          const message = request.getErrorMessage(body, error);
           return callback(`Requesting packages failed: ${message}`);
         }
       });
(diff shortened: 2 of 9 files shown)


confused-Techie:
We are seeing more failures of NodeJS 18 here, like we are elsewhere. So those shouldn't be considered during a final review of the PR.

DeeDeeG (approved):
Appears to work as intended with the `ppm star` command, which was the command reported as having not-working error handling on the Discord thread.

---

Current `master` branch code:

```console
% ./bin/ppm star does-not-exist
⭐  Starring does-not-exist ✗
Starring package failed: null
```

(Error message comes back as just `null`)

---

This branch:

```console
% ./bin/ppm star does-not-exist
⭐  Starring does-not-exist ✗
Starring package failed: Not Found
```

(Error message is a more useful "Not Found".)

confused-Techie:
Thanks a ton for the approval @DeeDeeG! I'll go ahead and merge this one

PR #103, opened 2023-12-13
src: Stop pinging backend during package uninstalls

DeeDeeG:
So, we stopped *using* these pings on our implementation of the backend, since it doesn't really make sense to decrement the install count when someone uninstalls... arguably. And it may or may not have simply been telemetry when the O.G. Atom backend was expecting this to be an authed action.

Since we don't need to auth or ping the backend for uninstalls... Just delete the code that does it!

### DIFF
--- a/src/uninstall.js
+++ b/src/uninstall.js
@@ -37,24 +37,6 @@ Delete the installed package(s) from the ~/.pulsar/packages directory.\
       }
     }
 
-    async registerUninstall({packageName, packageVersion}) {
-      if (!packageVersion) { return; }
-
-      try {
-        const token = await auth.getToken();
-        const requestOptions = {
-          url: `${config.getAtomPackagesUrl()}/${packageName}/versions/${packageVersion}/events/uninstall`,
-          json: true,
-          headers: {
-            authorization: token
-          }
-        };
-        return new Promise((resolve, _reject) => void request.post(requestOptions, (_error, _response, _body) => resolve()));
-      } catch (error) {
-        return error; // error as value here
-      }
-    }
-
     async run(options) {
       options = this.parseOptions(options.commandArgs);
       const packageNames = this.packageNamesFromArgv(options.argv);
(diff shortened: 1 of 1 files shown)


confused-Techie (approved):
Easy approval!
Thanks for finally getting caught up on this on, as it's been in the back of my head for far too long.

Any logic was [removed from the backend](https://github.com/pulsar-edit/package-backend/pull/105) way back in February, so this has certainly been a long time coming.

But for the most part tests look happy, beyond what can only be assumed external factors, so I'd say we are good to merge.

(Plus you were right, this is a really clean removal)

DeeDeeG:
> tests look happy, beyond what can only be assumed external factors

External factors, yes. Apparently we haven't ran CI on this repo in a while, so we weren't hit with the "older node-gyp, newer Python" issues here until now.

(GitHub Actions just pulled in the latest `macOS-latest` image, presumably with newer Python, on the CI runs of the second commit of this PR. Fist commit in this PR got an older runner image for its CI jobs, for whatever reason.)

ppm's bundled node-gyp is new enough to handle this, but the CI runner's Node 14 --> npm 6 --> node-gyp 5 is too old to work with Python 3.9+. Also, the CI runner's versions of npm --> node-gyp 9/older aren't patched for Python 3.12 yet either.

I'll see how we fixed this in other repos, but maybe that should be a separate PR to fix CI, since it's kind of unrelated to this PR, strictly speaking.

DeeDeeG:
Thank you for review, I'll go ahead and merge this, since it's tested working on my machine locally, works in CI (when CI isn't having a "node-gyp version" + "Python version" mutual incompatibility issue), etc.

Don't want to forget about this one or have to pick it up again tomorrow when I may have forgotten some of the details. It's fresh in my mind now, and I think it looks good, so here we go, merging.

src/uninstall.js
```diff
@@ -82,9 +61,6 @@ Delete the installed package(s) from the ~/.pulsar/packages directory.\
             if (fs.existsSync(packageManifestPath)) {
               const packageVersion = this.getPackageVersion(packageDirectory);
```
DeeDeeG, line 62: Whoops, didn't notice `packageVersion` is now defined but unused.

Didn't notice I could/should delete this line, too. Can do it in a follow-up, I guess.

PR #104, opened 2023-12-13
src: Delete unused code in uninstall.js

DeeDeeG:
This is a quick follow-up to some code removal in PR 103.

In #103, after removing some code that does a ping to the backend (which we no-op/ignore anyway, on the backend side...) I didn't notice `packageVersion` is now defined but unused with those code removals.

Basically this is one line I missed in the cleanup. Things that were part of making the backend ping work. With that removed, this line/variable serve no purpose.

This basically finishes the cleanup started in #103.

**(TL;DR: Another too-long writeup of a tiny diff. This is a one-liner removing a defined but unused var.)**

EDIT: Whoops, mentioned and tagged the wrong PR. This PR follows-up on PR 103, not PR 95...

### DIFF
--- a/src/uninstall.js
+++ b/src/uninstall.js
@@ -27,14 +27,6 @@ Delete the installed package(s) from the ~/.pulsar/packages directory.\
       return options.boolean('hard').describe('hard', 'Uninstall from ~/.pulsar/packages and ~/.pulsar/dev/packages');
     }
 
-    getPackageVersion(packageDirectory) {
-      try {
-        return CSON.readFileSync(path.join(packageDirectory, 'package.json'))?.version;
-      } catch (error) {
-        return null;
-      }
-    }
-
     async run(options) {
       options = this.parseOptions(options.commandArgs);
       const packageNames = this.packageNamesFromArgv(options.argv);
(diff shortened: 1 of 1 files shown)


DeeDeeG:
macOS CI failure is unrelated, it's the "old node-gyp, new Python" issues again.

I have to review how we worked around these things in CI for other repos. Wanted to get this PR out quickly, will have to come back around to fixing CI here soon.

DeeDeeG:
Wow, you're right, `getPackageVersion` can go too! This file has to be almost half the length it was before, with all this pruning.

confused-Techie:
@DeeDeeG Thanks for taking a look and confirming what I saw there.
More good news, now with that function removed, it looks like we can stop requiring `season` in this file as well!

DeeDeeG:
With that removal, I'm reasonably confident-ish  that there's no other dead/unused code that can simply be deleted from this file. Tempting fate with that, but I think this is good now(?).

confused-Techie (approved):
Sorry I kept adding onto this one.

But yup I agree that it seems everything has been taken from this file that we can for now.

In which case it looks good to me, and lets get it merged!

Thanks for putting this one together, and not minding me adding onto it

DeeDeeG:
No problem! Good to be thorough.

PR #106, opened 2023-12-13
Fix Newer NodeJS CI

confused-Techie:
This PR aims to resolve the CI issues we are seeing on this repo.

As activity has been spares over here, seems this repo had never yet received the new `setuptools` and Python 3.12+ actions logic.

So this PR simply takes what's mostly used over at Pulsar, and adds it's logic to the CI here.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -40,6 +40,14 @@ jobs:
         with:
           python-version: '3.12'
 
+      - name: Setup Python (NodeJS v14)
+        # While initially tests would pass with no Python setup, additional testing
+        # is showing issues with CI included Python versions, so we will install our own
+        if: ${{ matrix.node_version == 14 }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+
       - name: Install Python Dependencies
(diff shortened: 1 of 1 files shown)


DeeDeeG (approved):
Thank you for this!!!!!

Technically we can delete the "install Python 3.12" if we want, but this pins it so Python 3.13 (several months away) can't surprise us...

I slightly like to not pin stuff when I can get away with it, to make sure our stuff breaks in CI _first_ rather than have end users wonder why it didn't work for them. So we can catch it early in the repo and fix it early in the repo. But, not having your CI break for reasons outside of your code changes is a benefit of pinning, so.

Will accept this PR as-is, or changed to not manually install Python 3.12 (for the non Node 14 runs).

confused-Techie:
Thanks for the approval, tbh with the amount of issues Python has given us, I'd much rather add Python versions into the matrix if we wanted to run tests on version above what we know for a fact works.

But with everything else already pinned, I'm inclined to add as is, rather than deal with the surprises. But do totally get where you are coming from. But thanks for adding approval as is, which I'll merge, but am totally not against running tests on this Python version and newer (or latest) if we felt it was a good idea. Thanks!

PR #107, opened 2023-12-13
Repository Cleanup

confused-Techie:
Taking a closer look at this repo, now that the big ticket items have all been taken care of (eg Decaf, Async), I noticed quite a few outdated and useless files that had still remained.

I've gone ahead and removed the following files:
* `pnpm-workspace.yaml`: We do not use PNPM in this repo officially.
* `appveyor.yml`: This repo has fully switched over to GitHub Actions.
* `./github/no-response.yml`: We do not use probot.
* `npmignore`: We do not publish to NPM at all.
* `package-lock.json`: Since this repo has fully moved over to Yarn, this should no longer be needed.

The bonus of PRs like this is to simply get rid of clutter in the repo, allowing a cleaner broad picture of the codebase as a whole.


DeeDeeG (approved):
Looks good, a little cleanup is a good thing. I don't really expect us to be `npm pack`ing or `npm publish`ing this any time soon, so package-lock.json and .npmignore can go.

The other files are pure cruft, can't think of a situation where we'd miss them.

LGTM

PR #109, opened 2023-12-14
Prune outdated Deps

confused-Techie:
After all the recent modifications and updates in this repo, seemed to be a good idea to make sure we didn't have any old useless deps lying around.

Was able to find a few that are no longer in use:
* `glob`
* `q`
* `coffeelint`


DeeDeeG (approved):
Looks Good to me! (Tests passing.)

Linking to the other PR where we identified `coffeelint` should eventually be removed as a follow-up task at the time: https://github.com/pulsar-edit/ppm/pull/92

Great to prune these un-needed deps!

DeeDeeG:
Going to merge this one myself, very straightforward PR.

PR #112, opened 2023-12-15
Remove unused Variables

confused-Techie:
This PR removes most of the unused variables within the codebase. Advised via the Codacy recommendations.

This also has a focus on imports mostly, and ignored any unused variables that begin with `_` for the time being.

But for this PR while touching unrelated files, I hope keeping the purpose identical between all of them, it's not too hard to review.

### DIFF
--- a/src/clean.js
+++ b/src/clean.js
@@ -1,15 +1,8 @@
 
-let Clean;
-const path = require('path');
-
-const async = require('async');
-const CSON = require('season');
 const yargs = require('yargs');
-const _ = require('underscore-plus');
 
 const Command = require('./command');
 const config = require('./apm');
-const fs = require('./fs');
 
 module.exports =
 class Clean extends Command {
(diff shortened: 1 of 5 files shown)


DeeDeeG (approved):
Looks good to me!

CI passes, so generally the code still works. Missing imports would tend to error out early and loud, so I'm hoping CI is enough to catch these easily.

confused-Techie:
Thanks for the review @DeeDeeG!

DeeDeeG:
> ignored any unused variables that begin with _ for the time being

I was curious about these, but the few I found seem reasonable, and the convention about starting a var name with `_` to mean "yes, this is unused intentionally at the moment" is understandable to me at least. I can agree with leaving those as-is at the moment.

I'm honestly not as clear why those are ever super useful in JavaScript, and one or two of those aren't even unused, so it seems incorrect to start them with `_`, (`_gitName` and `_versionName` in `src/git.js`), but eh, off-topic maybe? Unless we want to rename those to not start with `_`, I'd be okay with doing that here, but it feels kind of tangential to the point of this PR, so I won't push for it really.

confused-Techie:
@DeeDeeG Merged this just as soon as you commented. But a followup PR to address any incorrect uses of `_` at the start of a variable name seems like an awesome idea.

But glad we are on the same page, it felt smart to leave these behind, since they are intentionally unused. As for the why, I feel like this convention comes up in most places where you are allowed to overload or underload the parameters of a function. Since leaving those unused values out could lead someone to misunderstand what parameters are expected, or could be available. I do see how it can make the API of some surfaces much easier to work with.

PR #116, opened 2023-12-16
Fix bugs found in `publish` after 1.112 release

savetheclocktower:
These are just the ones I found when trying to run `ppm publish patch`, but I'm sure there are other bugs. The PPM rewrite was gigantic.

### DIFF
--- a/src/apm.js
+++ b/src/apm.js
@@ -30,11 +30,11 @@ module.exports = {
       if (process.env.ATOM_RESOURCE_PATH) {
         return void process.nextTick(() => resolve(process.env.ATOM_RESOURCE_PATH));
       }
-  
+
       if (asarPath) { // already calculated
         return void process.nextTick(() => resolve(asarPath));
       }
-  
+
       let apmFolder = path.resolve(__dirname, '..');
       let appFolder = path.dirname(apmFolder);
       if ((path.basename(apmFolder) === 'ppm') && (path.basename(appFolder) === 'app')) {
@@ -43,7 +43,7 @@ module.exports = {
           return void process.nextTick(() => resolve(asarPath));
         }
       }
-  
+
       apmFolder = path.resolve(__dirname, '..', '..', '..');
       appFolder = path.dirname(apmFolder);
       if ((path.basename(apmFolder) === 'ppm') && (path.basename(appFolder) === 'app')) {
@@ -52,7 +52,7 @@ module.exports = {
           return void process.nextTick(() => resolve(asarPath));
         }
       }
-  
+
       switch (process.platform) {
         case 'darwin':
           return child_process.exec('mdfind "kMDItemCFBundleIdentifier == \'dev.pulsar-edit.pulsar\'"', (error, stdout, _stderr) => {
--- a/src/publish.js
+++ b/src/publish.js
@@ -66,7 +66,7 @@ have published it.\
       return new Promise((resolve, reject) => {
         this.fork(this.atomNpmPath, versionArgs, (code, stderr, stdout) => {
           stderr ??= '';
-          stdout ??= ''; 
+          stdout ??= '';
           if (code === 0) {
             this.logSuccess();
             resolve(stdout.trim());
@@ -140,7 +140,7 @@ have published it.\
     //
     // packageName - The string package name to check.
     //
-    // return value - A Promise that can reject with an error or resolve to 
+    // return value - A Promise that can reject with an error or resolve to
     //                a boolean value.
     async packageExists(packageName) {
       const token = await Login.getTokenOrLogin();
@@ -186,7 +186,7 @@ have published it.\
 
       try {
         const token = await Login.getTokenOrLogin();
-        
+
         const requestSettings = {
           url: config.getAtomPackagesUrl(),
           json: true,
@@ -226,7 +226,7 @@ have published it.\
     //
     // return value - A Promise that rejects with an error or resolves without a value.
     async createPackageVersion(packageName, tag, options) {
-      const token = await Login.getTokenOrLogin();        
+      const token = await Login.getTokenOrLogin();
       const requestSettings = {
         url: `${config.getAtomPackagesUrl()}/${packageName}/versions`,
         json: true,
@@ -381,7 +381,7 @@ have published it.\
       let packageName, semverRange;
       if (!pack) { return; }
 
-      const isValidRange = function(semverRange) {
+      const isValidRange = function (semverRange) {
         if (semver.validRange(semverRange)) { return true; }
 
         try {
@@ -462,7 +462,7 @@ have published it.\
           rename = pack.name;
           pack.name = originalName;
         }
-        
+
         try {
           await this.publishPackage(pack, tag, {rename});
         } catch (error) {
@@ -478,7 +478,7 @@ have published it.\
         } catch (error) {
           return error;
         }
-        
+
         try {
           await this.publishPackage(pack, tag);
         } catch (error) {
--- a/src/publish.js
+++ b/src/publish.js
@@ -114,20 +114,25 @@ have published it.\
       };
 
       return new Promise((resolve, _reject) => {
-        const requestTags = () => void request.get(requestSettings, (_error, response, tags) => {
-          tags ??= [];
-          if (response?.statusCode === 200) {
-            if (tags.find(elem => elem.name === tag) != null) {
-              resolve();
-              return;
+        const requestTags = () => {
+          request.get(
+            requestSettings,
+            (_error, response, tags) => {
+              tags ??= [];
+              if (response?.statusCode === 200) {
+                if (tags.some(t => t.name === tag)) {
+                  resolve();
+                  return;
+                }
+              }
+              if (--retryCount <= 0) {
+                return void resolve();
+              }
             }
-          }
-          if (--retryCount <= 0) {
-            return void resolve();
-          }
-
+          )
           setTimeout(requestTags, interval);
-        });
+        };
+        requestTags();
       });
     }
 
@@ -169,7 +174,7 @@ have published it.\
       }
 
       const exists = await this.packageExists(pack.name);
-      if (exists) { return Promise.reject(); }
+      if (exists) { return Promise.resolve(false); }
 
       const repository = Packages.getRepository(pack);
 
@@ -324,7 +329,7 @@ have published it.\
 
     // Rename package if necessary
     async renamePackage(pack, name) {
-      if (name?.length <= 0) {
+      if ((name ?? '').length <= 0) {
         // Just fall through if the name is empty
         return; // error or return value?
       }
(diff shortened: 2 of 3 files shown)


savetheclocktower:
Added some tests. I don't love them — they construct a false reality! — but they're better than the nothing that we had.

asiloisad:
I have catch the same bug

image

savetheclocktower:
These new tests pass locally for me. I'll get around to fixing them.

savetheclocktower:
OK, found the problem with the tests. CI is good. This one is definitely a candidate for “squash and merge.”

Daeraxa:
Do we need a 1.112.1 for this?

savetheclocktower:
Yes. We should test other aspects of `ppm` to see if other bugfixes can be addressed in a patch release, but so far the only bugs reported are with publishing.

src/apm.js
DeeDeeG, line 1: Good to get rid of errant whitespace, but I would prefer to do it in a commit that outright says "get rid of excess whitespace" or "fix whitespace" for the sake of readable commit logs.

So, I can either fix up the branch to have such a commit, or revert the changes to this file, since it's whitespace-only, it's not part of the bug fixes.

src/publish.js
```diff
@@ -66,7 +66,7 @@ have published it.\
       return new Promise((resolve, reject) => {
         this.fork(this.atomNpmPath, versionArgs, (code, stderr, stdout) => {
           stderr ??= '';
-          stdout ??= ''; 
+          stdout ??= '';
```
DeeDeeG, line 69: The whitespace here was wrong before, but it's arguably slightly noisy changing this without it being in a commit that says "fix whitespace" or some such label.

src/publish.js
```diff
@@ -181,7 +185,7 @@ have published it.\
 
       try {
         const token = await Login.getTokenOrLogin();
-        
+
```
DeeDeeG, line 188: Whitespace-only change again

src/publish.js
```diff
@@ -221,7 +225,7 @@ have published it.\
     //
     // return value - A Promise that rejects with an error or resolves without a value.
     async createPackageVersion(packageName, tag, options) {
-      const token = await Login.getTokenOrLogin();        
+      const token = await Login.getTokenOrLogin();
```
DeeDeeG, line 228: whitespace-only change again

src/publish.js
```diff
@@ -456,7 +461,7 @@ have published it.\
           rename = pack.name;
           pack.name = originalName;
         }
-        
+
```
DeeDeeG, line 464: whitespace-only change

src/publish.js
```diff
@@ -472,7 +477,7 @@ have published it.\
         } catch (error) {
           return error;
         }
-        
+
```
DeeDeeG, line 480: whitespace-only change

DeeDeeG (commented):
Putting all the whitespace comments I had in their own thread so it can easily be minimized when looking at the actual substantive feedback.

src/apm.js
savetheclocktower: Sorry, I've got this weird editor called Pulsar that strips whitespace from the ends of lines when I save.

src/publish.js
```diff
@@ -169,7 +173,7 @@ have published it.\
       }
 
       const exists = await this.packageExists(pack.name);
-      if (exists) { return Promise.reject(); }
+      if (exists) { return Promise.resolve(false); }
```
DeeDeeG, line 176: This change looks good 👍

<details><summary>Context from looking and the pre-asyncified code (click to expand):</summary>

See the old code here: 

https://github.com/pulsar-edit/ppm/blob/9b7205b04b823fd4835b97fef08c467936541bc2/src/publish.js#L169

We can see that, before the "`async`ification" PR, this function did not throw an error here if `exists` was truthy, rather it just returned the callback without no argument -- i.e. without a truthy argument. Returning `false` here without erroring is fairly equivalent to pre-`async`ification.

---

</details>

**I suppose we need to update the comment above:**

`or resolve with true value` --> `or resolve with true/false value` (or something like that)

DeeDeeG (commented):
I was able to reduce the diff in `src/` dir to the following and pass all the new tests:

```diff
diff --git a/src/publish.js b/src/publish.js
index 4dfef93..4571e44 100644
--- a/src/publish.js
+++ b/src/publish.js
@@ -169,7 +169,7 @@ have published it.\
       }
 
       const exists = await this.packageExists(pack.name);
-      if (exists) { return Promise.reject(); }
+      if (exists) { return false; }
 
       const repository = Packages.getRepository(pack);
 
@@ -324,7 +324,7 @@ have published it.\
 
     // Rename package if necessary
     async renamePackage(pack, name) {
-      if (name?.length <= 0) {
+      if ((name ?? '').length <= 0) {
         // Just fall through if the name is empty
         return; // error or return value?
       }
```

Along with updating one comment to reflect `return false` instead of `Promise.reject()` as a return value, is it okay to ask that the PR be reduced to this change? And I can volunteer to do this if the PR author is okay with my doing so and prefers not to spend the time fiddling with it themselves.

Open to feedback as to no reducing the `src/` changes in the PR to the above, though.

src/publish.js
```diff
@@ -114,28 +114,32 @@ have published it.\
...
-            return void resolve();
-          }
-
-          setTimeout(requestTags, interval);
-        });
+          )
+        };
+        setTimeout(requestTags, interval);
       });
     }
```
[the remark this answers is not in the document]
savetheclocktower: I rewrote this section because the original code (in my opinion) went out of its way to be inscrutable.

An arrow function can skip surrounding braces if the entire body is a single statement. That doesn't mean it's a good idea to treat a complex multi-line statement as though it were a one-liner. I think that this is one of the reasons why the author ended up putting the `setTimeout` on the inside of the `requestTags` callback instead of the outside.

src/publish.js
savetheclocktower: I suppose it could be an ordinary `return false`. I don't have an opinion on which one is better.

src/publish.js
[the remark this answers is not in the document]
savetheclocktower: I just realized you're right that I shouldn't have moved the `setTimeout` — I should've just invoked `requestTags` once outside the block. I'll make that change.

I'm still strongly in favor of reformatting that code; I'd prefer to do it now rather than in a separate PR.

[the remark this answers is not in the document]
savetheclocktower: And you're right that this code path still isn't hit in the tests — even in the new tests I wrote. I mocked anything that talked to the outside world. I don't know what else to do for a method whose only purpose is to wait until a remote server responds a certain way.

src/publish.js
```diff
@@ -440,10 +445,11 @@ have published it.\
         if (rename?.length > 0) { originalName = pack.name; }
 
         let firstTimePublishing;
+        let tag;
```
savetheclocktower, line 448: Along with the two changes you mention in your review summary, this change is crucial. It moves `tag` to the outer scope so that it can be consumed later on.

src/publish.js
```diff
@@ -114,28 +114,33 @@ have published it.\
...
-          }
-          if (--retryCount <= 0) {
-            return void resolve();
-          }
-
+          )
           setTimeout(requestTags, interval);
-        });
+        };
+        requestTags();
```
savetheclocktower, line 135: And this is the fourth change: invoking `requestTags` immediately after defining it. Without this, you can see that `requestTags` never gets called in the first place.

These four changes need to be part of any fix. I would also like this formatting change to make it in.

I understand what you're saying about the whitespace. If you're saying that all whitespace-related changes should go into their own commit, that's fine. If you're saying that they shouldn't make it into this PR at all, that'd be harder for me to agree with. But it's not something I'd go twelve rounds over.

src/publish.js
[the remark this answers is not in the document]
DeeDeeG: Re: **tests**, yeah, our tests are already not platinum-tier 110% complete, and I don't know off the top of my head how to mock an error code response from the hypothetical server. Certainly not going to ask for this, it's not obvious to me how doable it would be.

As for readability, yeah the original syntax was a little... weird. It's still kinda gnarly due to being so nested and having so many different JS features used in such a short span. **I think you're right that a refactor is warranted for readability.**

**I ended up with this locally just so I could read and understand what was happening.**

<details><summary>Code snippet (click to expand):</summary>

```js
    waitForTagToBeAvailable(pack, tag) {
      let retryCount = 5;
      const interval = 1000;
      const requestSettings = {
        url: `https://api.github.com/repos/${Packages.getRepository(pack)}/tags`,
        json: true
      };

      return new Promise((resolve, _reject) => {
        const requestTags = () => void request.get(requestSettings, (_error, response, tags) => {
          tags ??= [];
          if (response?.statusCode === 200) {
            if (tags.find(elem => elem.name === tag) != null) {
              resolve();
              return;
            }
          }
          if (--retryCount <= 0) {
            return void resolve();
          }

          setTimeout(requestTags, interval);
        });
      });
    }
```

</details>

Took me a lot of staring at it and having Pulsar highlight matching brackets/parentheses for me and trying to run it to see which line errored...

DeeDeeG (approved):
Looks good to me.

I am planning to manually rebase this branch so it's a cleaner, simpler series of commits, with any whitespace-only lines of diff in their own commit or commits as appropriate.

If you want to do the whitespace stuff or the rebase and cleanup, you can, but I was going to do it myself rather than make you worry about stuff outside what the code should look like.

DeeDeeG:
Okay, I saved a copy of the original branch at https://github.com/DeeDeeG/ppm/tree/bug-fixes-original, for reference.

Branch protection rules won't let me push `master` branch without it coming from a PR, so I'll update (force push) this PR with the tidied up commit history real quick...

DeeDeeG:
Oh, actually... Hmm. I think I accidentally bypassed the branch protection rules (that I didn't recall we had set up) with admin powers.

DeeDeeG:
Manually merged via commit https://github.com/pulsar-edit/ppm/commit/f660a7a70d9ce0b7b5f7c0fe1e2af6890af205fa (a fake-GitHub-merge-commit-looking actual-merge-commit that I prepared on the CLI).

DeeDeeG:
So, if you prefer the PR to look purple and say "merged" I'll back out the manual merge and do it via the GUI here, but otherwise this'll be red and "closed", but the changes are in `master` branch at the moment one way or the other.

confused-Techie:
@DeeDeeG For tracking and changelog reasons I'd say lets properly merge this one, but at the end of the day, as long as we remember the changes are in here, I suppose it doesn't matter too much

DeeDeeG:
CI passed with my rebased commits (as they should, I didn't change any code at all), and I can confirm the "diff" between the original and the rebased version is none, only thing I did was the cleaner commit history.

Finally merging. Thanks much for the fix and added tests!

PR #118, opened 2023-12-19
Fix placement of `setTimeout`

savetheclocktower:
The command was hanging because `setTimeout` was being called over and over, rather than just as needed (because the tag wasn't available yet). The `setTimeout` started out in one wrong place and then I “fixed” it by moving the `setTimeout` to a _different_ wrong place.

There is a separate issue here: the API calls to GitHub could fail for any number of reasons. One reason they failed for me was because I exceeded the unauthenticated rate limit! This command should at least handle the error cases, but I believe it can also use the very auth token we use for the package backend in order to enable a much higher rate limit. Both of those are worth doing, but I'll save them for a different PR.

### DIFF
--- a/src/publish.js
+++ b/src/publish.js
@@ -128,9 +128,9 @@ have published it.\
               if (--retryCount <= 0) {
                 return void resolve();
               }
+              setTimeout(requestTags, interval);
             }
-          )
-          setTimeout(requestTags, interval);
+          );
         };
         requestTags();
       });


DeeDeeG (approved):
Tested working, thank you very much!

To clarify the issue / expected:

Issue: `ppm publish` succeeds but never exits.

Expected: `ppm publish` should succeed and exit, returning the CLI to the user's control for further commands.

Fix: Allows `ppm publish` to exit as it should after it finishes.


Main test method: `ppm publish patch`.

Also tried some invalid things to have `ppm publish` exit with errors, and that worked as it should too.

PR #120, opened 2023-12-22
Removing "No Operand in Assignment" Usage

confused-Techie:
With PPM now properly setup on Codacy, the next two biggest suggestions are the following:

* No Unused Vars: For this one I've added a small bit of configuration to the eslint config file, to not apply this rule if the variable begins with an underscore, which should solve solve that issue by and large for us.
* No Assignment in Operand: This issue is less invasive than it appears. While there are some instances where variables are assigned a value within an operand, other situations were found such as old CoffeeScript `is not` operators that weren't decaffed cleanly, and also a single instance of what appears to be a typo of `=` rather than `==`

With this issues addressed, we don't expect to see any change in hope behavior ourselves.

### DIFF
--- a/src/list.js
+++ b/src/list.js
@@ -106,15 +106,14 @@ List all the installed packages and also the packages bundled with Atom.\
     listPackages(directoryPath, options) {
       const packages = [];
       for (let child of Array.from(fs.list(directoryPath))) {
-        var manifestPath;
         if (!fs.isDirectorySync(path.join(directoryPath, child))) { continue; }
         if (child.match(/^\./)) { continue; }
         if (!options.argv.links) {
           if (fs.isSymbolicLinkSync(path.join(directoryPath, child))) { continue; }
         }
 
-        let manifest = null;
-        if (manifestPath = CSON.resolve(path.join(directoryPath, child, 'package'))) {
+        let manifestPath = CSON.resolve(path.join(directoryPath, child, 'package'));
+        if (manifestPath) {
           try {
             manifest = CSON.readFileSync(manifestPath);
           } catch (error) {}
--- a/src/list.js
+++ b/src/list.js
@@ -112,6 +112,7 @@ List all the installed packages and also the packages bundled with Atom.\
           if (fs.isSymbolicLinkSync(path.join(directoryPath, child))) { continue; }
         }
 
+        let manifest = null;
         let manifestPath = CSON.resolve(path.join(directoryPath, child, 'package'));
         if (manifestPath) {
           try {
--- a/src/publish.js
+++ b/src/publish.js
@@ -314,7 +314,7 @@ have published it.\
       }
 
 
-      if (currentBranch = repo.getShortHead()) {
+      if (currentBranch == repo.getShortHead()) {
         remoteName = repo.getConfigValue(`branch.${currentBranch}.remote`);
       }
       if (remoteName == null) { remoteName = repo.getConfigValue('branch.master.remote'); }
--- a/src/publish.js
+++ b/src/publish.js
@@ -314,7 +314,8 @@ have published it.\
       }
 
 
-      if (currentBranch == repo.getShortHead()) {
+      currentBranch = repo.getShortHead();
+      if (currentBranch) {
         remoteName = repo.getConfigValue(`branch.${currentBranch}.remote`);
       }
       if (remoteName == null) { remoteName = repo.getConfigValue('branch.master.remote'); }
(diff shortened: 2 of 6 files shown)


src/publish.js
```diff
@@ -314,7 +314,7 @@ have published it.\
       }
 
 
-      if (currentBranch = repo.getShortHead()) {
+      if (currentBranch == repo.getShortHead()) {
```
DeeDeeG, line 317: Suggesting this code change, since I think this is meant to be assignment after all. So, it looks like we can handle it like all the other "no assignment in operand" lints:

```suggestion
      currentBranch = repo.getShortHead();
      if (currentBranch) {
```

---

<details><summary>Long version / breakdown / context (click to expand):</summary>

Looks like this was actually meant to be an assignment rather than a comparison after all. It was assignment back in the CoffeeScript: 

https://github.com/pulsar-edit/ppm/blob/de1221ea743721bb68cf895596528f7a2251a7f9/src/publish.coffee#L252-L253

I think what's happening here is, based on the [`git-utils` documentation](https://github.com/atom/git-utils#readme), `git-utils` generally seems to prefer returning `null` rather than throwing Errors if something weird happens under the hood (?). So you can basically check if something from `git-utils` worked by checking if its result was truthy or null.

So, we could expect the result of `repo.getShortHead()` to be a shortened SHA (string, truthy), or null (falsy). And the `if ()` check is seeing whether the `repo.getShortHead()` succeeded or not/whether we have an actual string value to work with or not.

</details>

I definitely agree with separating out these sorts of things into more lines, so the linter warning about this makes sense.

DeeDeeG (commented):
Overall looking good, I think this one change is meant to be assignment not comparison, though.

Thanks for doing this, IMO these complicated lines really do make more sense separated into more lines.

src/list.js
```diff
@@ -106,15 +106,14 @@ List all the installed packages and also the packages bundled with Atom.\
...
       const packages = [];
       for (let child of Array.from(fs.list(directoryPath))) {
-        var manifestPath;
         if (!fs.isDirectorySync(path.join(directoryPath, child))) { continue; }
         if (child.match(/^\./)) { continue; }
         if (!options.argv.links) {
           if (fs.isSymbolicLinkSync(path.join(directoryPath, child))) { continue; }
         }
 
-        let manifest = null;
```
DeeDeeG, line 116: Might have been deleted by accident, `manifest` is checked later in the function.

We could probably refactor so it's not checked later, if we want, but for now we can restore this line `let manifest = null;`

src/publish.js
confused-Techie: Thanks for doing the research into this one! Fantastic explanation, and seems you are 100% correct

src/list.js
confused-Techie: Thanks for calling this one out, restored that line, not sure what went wrong there

src/upgrade.js
```diff
@@ -52,8 +52,8 @@ available updates.\
     getInstalledPackages(options) {
       let packages = [];
       for (let name of fs.list(this.atomPackagesDirectory)) {
-        var pack;
-        if (pack = this.getIntalledPackage(name)) {
+        let pack = this.getInstalledPackage(name);
```
DeeDeeG, line 55: You're probably not going to believe this, but the actual function name in this file has been `getIntalledPackage` (note the typo, a missing "s" in "intalled") this whole time, since 2014 (https://github.com/atom/apm/commit/c8d61000#diff-0aa64babcef87ba238fed061035f2fe14c587092a7339b55c5e4033c37370d70R46).

(Regarding the singular `getIntalledPackage`, not to be confused with the plural `getInstalledPackages` ("Package**s**") which has/had no typo.)

So, we can either revert the spelling here to the previous (typoed but defined) function name here, or rename the function elsewhere in the file that this line calls, so as to not have a typo in its name, lol.

But yeah, this broke tests by fixing a typo, I honestly did not see that coming.

Thanks again for working on this! Nice to see readability improvements to the code!!!!!! We sure need it, as debugging this PR has helped prove, IMO!!! Thank you!

DeeDeeG, line 55:
```suggestion
        let pack = this.getIntalledPackage(name);
```

This makes tests pass again for me.

confused-Techie: Thanks for the diff @DeeDeeG, and thanks for staying more dedicated to my PR than I've apparently been lol. I'll get this one merged and at a later time deal with the typo

DeeDeeG: No worries, it keeps showing in my GitHub inbox, that's how I keep noticing it's pending yet. Very close to the finish line so I thought I'd bop in and post a diff.

DeeDeeG (approved):
_Yeeeeeeeeeaaaaaaaaaah 😎_

let's do it

(More readable code that at least mostly makes sense is both a good thing and has historically been all too rare in ppm repo. We've come a long way!!!!!!!)

PR #124, opened 2024-01-29
Fix test failure due to missing atom command

toddy15:
If pulsar is installed from the .deb package,
the command `ppm test` fails with this error:

`Error spawning atom: spawn atom ENOENT`

Using `pulsar` instead of `atom` fixes this.

### DIFF
--- a/spec/test-spec.js
+++ b/spec/test-spec.js
@@ -30,8 +30,8 @@ describe('apm test', () => {
     waitsFor('waiting for test to complete', () => atomSpawn.callCount === 1);
 
     runs(() => {
-      // On Windows, there's a suffix (atom.cmd), so we only check that atom is _included_ in the path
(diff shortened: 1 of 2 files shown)


toddy15:
I did not update `spec/test-spec.js`, this is fixed now. Hopefully all tests should now pass.

DeeDeeG (approved):
Thanks for this fix! (And for updating the associated test file.)

Looks good to me 👍

DeeDeeG:
Just popping in to suggest we might shoot for two of these PRs per month? May mess with your PR titles if we do that, maybe Feb part 1, Feb Part 2? I dunno. In any case, I am thankful for this continued work.

DeeDeeG:
CI is passing, thanks again! Merging now. 👍

PR #132, opened 2024-05-18
Rewrite publish logic

confused-Techie:
After numerous issues and reports and pain points of this publish logic, we hit a point after enabling strict versioning on the backend to discover that the original logic completely fails now.

I'll try to update this with more details with more time, but for now, here's an attempt that should resolve our woes.

### DIFF
--- a/src/publish.js
+++ b/src/publish.js
@@ -424,13 +424,17 @@ have published it.\
       let {tag, rename, verbose} = options.argv;
       let [version] = options.argv._;
 
+      // Normalize variables to ensure they are strings with zero length
+      // rather than their default `undefined`
+      // This ensures later length checks always behave as expected
       if (typeof tag !== "string") {
-        // prevent tag from being undefined, and make it a string with 0 length
         tag = "";
       }
       if (typeof version !== "string") {
-        // prevent tag from being undefined, and make it a string with 0 length
         version = "";
+      }
+      if (typeof rename !== "string") {
+        rename = "";
       }
 
       if (verbose) {
--- a/src/publish.js
+++ b/src/publish.js
@@ -453,13 +453,13 @@ have published it.\
         return error;
       }
 
-      if (version?.length === 0 && tag?.length === 0 && rename?.length === 0) {
+      if (version.length === 0 && tag.length === 0 && rename.length === 0) {
         return "A version, tag, or new package name is required";
       }
 
-      if (rename?.length > 0) {
+      if (rename.length > 0) {
         // A version isn't required when renaming a package (apparently)
-        if (version?.length === 0) { version = "patch"; }
+        if (version.length === 0) { version = "patch"; }
         originalName = pack.name;
 
         try {
@@ -472,7 +472,7 @@ have published it.\
       // Now we know a version has been specified, and that we have settled any
       // rename concerns. Lets get to publication
 
-      if (tag?.length === 0) {
+      if (tag.length === 0) {
         // only create and assign a tag if the user didn't specify one.
         // if they did we assume that tag already exists and only work to publish
         try {
--- a/src/publish.js
+++ b/src/publish.js
@@ -359,7 +359,7 @@ have published it.\
           this.spawn(gitCommand, ['commit', '-m', message], (code, stderr, stdout) => {
             stderr ??= '';
             stdout ??= '';
-            if (code === 0) {
+            if (code !== 0) {
               this.logFailure();
               const commitOutput = `${stdout}\n${stderr}`.trim();
               reject(`Failed to commit package.json: ${commitOutput}`);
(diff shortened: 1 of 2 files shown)


DeeDeeG:
Posting this feedback from the Discord thread, in case it gets lost otherwise:

---

`ppm publish --tag` isn't working with this branch:
```
% ~/ppm/bin/ppm publish --tag v0.3.0
A version, tag, or new package name is required
```
Working on `master` branch:
```
% ~/ppm/bin/ppm publish --tag v0.3.0
Publishing deedeeg-my-package-10982367019823@v0.3.0 ✓
```

Sadly it's seeming like we might not have any unit tests for this?

DeeDeeG:
Speaking of which, the Discord thread: https://discord.com/channels/992103415163396136/1241217301529886853

And the start of the recent round of discussion around this topic, preceding the formal Discord thread about it by a bit: https://discord.com/channels/992103415163396136/992109539346370661/1240836443299905606

src/publish.js
```diff
@@ -440,56 +453,71 @@ have published it.\
         return error;
       }
 
-      if ((version?.length > 0) || (rename?.length > 0)) {
-        let originalName;
-        if (version?.length <= 0) { version = 'patch'; }
-        if (rename?.length > 0) { originalName = pack.name; }
+      if (version?.length === 0 && tag?.length === 0 && rename?.length === 0) {
```
DeeDeeG, line 456: I think it would be a benefit to know this is somehow still ending up `undefined` (despite the normalization we're doing) sooner, due `undefined.length` giving an error `Uncaught TypeError: Cannot read properties of undefined (reading 'length')`.

```suggestion
      if (version.length === 0 && tag.length === 0 && rename.length === 0) {
```

Sort of theoretical, since it shouldn't be possible.

But also, I like to avoid more complicated syntax if it's not needed, for ease of reading.

Also: Given the head-scratcher this issue was to solve, correctness and surfacing error messages in as straight-forward a way as possible is kind of my top priority, rather than avoiding errors in this case. Unless we want to try/catch and print the error and exit gracefully, but... I don't think that helps anything anyway in this case.

**tl;dr: Hitting a simple error message would be better than not hitting it, in this case?**

Not a blocker, since it should all be hypothetical with all the normalizing done in this PR. But the goose-chase tracking down strange behavior makes me a tad paranoid, not gonna lie.

DeeDeeG: Sorry for all the nitpicking, just inclined to make this code _very tidy_ if possible after what had to be thought through to get here.

DeeDeeG: _Note: If moving away from optional chaining for the `varName.length` checks, there are three more in the diff for this PR, I just didn't individually comment them all._

DeeDeeG: P.S. I suppose these instances of optional chaining mirror what was there before, but while we're addressing tiny subtleties of behavior that were unintuitive, I think it's worth deciding whether to move away from the optional chaining or not.

DeeDeeG (approved):
Giving the ol' "Approve with optional feedback".

Looks Good to Me, tested _thoroughly_ in the linked Discord thread. Worked for both of us on separate machines, should be in good shape now. 👍

src/publish.js
```diff
@@ -359,7 +359,7 @@ have published it.\
           this.spawn(gitCommand, ['commit', '-m', message], (code, stderr, stdout) => {
             stderr ??= '';
             stdout ??= '';
-            if (code === 0) {
+            if (code !== 0) {
```
DeeDeeG, line 362: Kind of odd this was wrong, I believe it was inadvertently changed as part of the _utterly massive_ promisification PRs, from looking at the blame history.

This change seems legit. All other calls to `logFailure()` are when `code !== 0`, as in error code rather than clean exit code. So yeah, seems legit IMO.

DeeDeeG (approved):
Approving, since this fixes the various intended use-cases, such as `ppm publish patch` for a first publish, while keeping `--tag` working, and fixing `--rename` which may have been broken for some time on `master` branch.

Might be a good idea either in this PR to do a follow-up to handle invalid combos of CLI args, such as doing `patch`, `--tag`, and `--rename` all at once? But I'm not sure that's actually invalid or if someone could intentionally do something useful with combining them? So, this is _optional_ to address, IMO. We can tell people to do just one at a time without having the code print an error message about it for now?

If you do limit to only one of them being set at a time, that could be reasonable, but honestly I think it's good enough for now without that as well.

src/publish.js
confused-Techie: Was surprised to see this one broken for so long tbh

confused-Techie:
Thanks for the approval, gonna go ahead and merge this one for now, then we just gotta get it updated in pulsar sometime soon

PR #141, opened 2024-09-09
Remove hard-coded architecture on Mac

### LINKED ISSUE #126: Add ability for PPM to run `electron-rebuild` automatically
### Have you checked for existing feature requests?

- [X] Completed

### Summary

Atom occasionally needed to rebuild the dependencies of community packages. For one thing, some NPM packages don't have prebuilds for some platforms. And when Atom upgraded its own version of Electron, that would require a rebuild of the native modules of previously installed community packages so that they could work with the new Electron version.

It _seems_ that this used to be possible simply by running `ppm rebuild [package]`. That command delegates to `npm rebuild` (using `ppm`'s internal copy of `npm`, of course). And Pulsar's builtin `incompatible-packages` package offers a GUI for performing this task — which internally just shells out to `ppm rebuild`.

Nowadays, however, this doesn't seem to be enough; these tasks claim to work, but don't produce modules capable of working in Pulsar. One possible fix for this problem would be to rewrite PPM's `rebuild` task to delegate to `electron-rebuild`.

This will require two things:

1. We need a way to find the correct version of Electron for the currently installed Pulsar instance. We could just shell out to `pulsar` the way that the `test` tasks seems to do, taking care to ensure we're finding the version of the command that pairs with the version of `ppm` that's currently running. (In other words, don't just run `pulsar` and rely on `PATH` to find it.)
2. We need `electron-rebuild` to be a dependency of `ppm`. Currently it's only a dev dependency of Pulsar itself, while PPR doesn't require or use it at all.

I'm inclined to assume control of the existing `rebuild` task here because I don't _think_ that it does anything that `electron-rebuild` won't be able to do, but I'm happy to discuss that if anyone is concerned about it.

The short-term goal isn't to do anything magical or fancy with this — just to make it so that `ppm rebuild [package]` actually does what it says on the tin, instead of being seemingly useless. If this works, we can have a discussion about whether this should be a mandatory post-install task for a given package. If `electron-rebuild` understands when it does _not_ need to rebuild a module, then I don't see a downside to doing this, but we don't need to decide it now.

(**Reminder to myself or someone else:** before working on this, I know I need to understand the installation process better, and exactly what happens when `ppm` installs a package that uses native modules. For instance, I've never need to run `electron-rebuild` on any package that uses `@savetheclocktower/atom-languageclient`, even though it has a dependency with a native module. Why is my experience different from that of @mjrodgers (as reported in Discord)? He's on ARM64 and I'm on Intel, but surely a native module would need a rebuild anyway for an Electron environment, right? I should try to run (e.g.) `ppm install --verbose pulsar-ide-typescript-alpha` in a clean environment and pore over the output to see if I can answer these questions.)

### What benefits does this feature provide?

Easier package installations.

### Any alternatives?

@mauricioszabo wants a fix that doesn't depend on PPM, but that can be pursued in parallel.

### Other examples:

_No response_

savetheclocktower:
Fixes #126.

We've known for a while about an issue where `ppm rebuild` fails to work properly on Apple Silicon. That command is run whenever a native module needs to be recompiled — either for a new processor architecture or for a new version of Electron — and there's even a GUI for it.

This is what it might look like for an end user, as it did for me last week when I got a new M3 MacBook Pro:

https://github.com/user-attachments/assets/77b27b32-2937-4b61-a82e-bba419c08993

The built-in `update-package-dependencies` package is a frontend to `ppm rebuild`. Notice how it doesn't say “failed to rebuild the package”; it thinks it's succeeded, and yet after a window reload it once again flags the native module as being for the wrong architecture.

This was on my radar because some of `pulsar-ide-python`’s users had complained about it. `zadeh` was used in `atom-languageclient` — I moved away from it in my `@savetheclocktower/atom-languageclient` fork in order to work around this issue — so it affected a number of packages. It was strange to learn, then, that [`npm rebuild` seemed not to be susceptible to this issue](https://github.com/pulsar-edit/ppm/issues/126#issuecomment-1960101764). This reduced it to an exercise in finding out exactly what the discrepancy was between `ppm rebuild` and `npm rebuild`.

For a few hours I was certain it was some sort of artifact of a too-old version of Node, NPM, or `node-gyp`. I noticed a behavior difference between two of my local versions of Node — `14.16.0` did the wrong thing, whereas `16.19.0` did the right thing — and I generally wasted my own time until I looked closely enough at the logging and traced the issue back to the PPM source itself.

PPM provides a bunch of build flags for NPM, processor architecture among them; on `darwin` it's been hard-coding `x64` for [years and years and years](https://github.com/pulsar-edit/ppm/commit/7d923b2d0b2997056d60409814c3cde05413480b). I don't know why, but it was implemented in such a way that you couldn't even override it with an environment variable.

Removing the hard-coded special case seems to fix it for me. We should ensure it works as expected on an Intel Mac, but this feels like the obvious obvious obvious fix. I'm glad that it's as simple as this, though somewhat annoyed that I didn't notice it earlier.

Since neither `ppm rebuild` nor `npm rebuild` seems to need a subsequent pass through `electron-rebuild`, this should fix #126 altogether. ~~(This might even make it so that (e.g.) `ppm install autocomplete-paths` installs the package properly the first time around; I'll check that next.)~~ (Yeah, seems to work.)

I hope we can get this into the next release as a quality-of-life issue, but it can wait a month if people are nervous about it.

### DIFF
--- a/src/apm.js
+++ b/src/apm.js
@@ -95,12 +95,7 @@ module.exports = {
   },
 
   getElectronArch() {
-    switch (process.platform) {
-      case 'darwin':
-        return 'x64';
-      default:
-        return process.env.ATOM_ARCH ?? process.arch;
-    }
+    return process.env.ATOM_ARCH ?? process.arch;
   },
 
   getUserConfigPath() {


DeeDeeG:
Commenting for posterity, since I already went through the trouble of tracing back in time through the blame view fully (following your lead), and posted it in the Discord... copy-pasting here for the historical record.

> So, this code was from https://github.com/atom/apm/pull/46 which... dubiously introduced the hard-coding, [`a9e5498`](https://github.com/atom/apm/commit/a9e5498a838bf228fca53c32f66e2e6d5adcf783) which fixed it for Linux/BSD (by not hard-coding on Linux or BSD), and then https://github.com/atom/apm/pull/640 which fixed it for Windows by not hard-coding arch for Windows... and now we're not hard-coding it on macOS.

savetheclocktower:
> I noticed a behavior difference between two of my local versions of Node — `14.16.0` did the wrong thing, whereas `16.19.0` did the right thing

Oh, I forgot to mention: this is because I use `asdf` to manage my Node installations, and my `14.16.10` was built for x86_64. I had reinstalled most of my Nodes after migration, but not that one. (Just as well because there are no official arm64 builds of Node 14.)

DeeDeeG (approved):
tl;dr: FIX LOOKS GOOD TO ME. 👍 

Long version:

Can confirm that, on my x64 (intel) MacBook, Node returns `x64` when evaluating `process.arch`. Which matches what the hard-coded string would have been from the `switch`/`case` there was before this PR.

From that, and judging by the fact you're reporting this works for you on ARM (Apple Silicon), behavior should absolutely be correct across x64 and ARM Macs alike now.

Given the history of un-hard-coding these values to solve issues over time, this fix falls along a long and steady line of the same fix being implemented individually for every major OS family. It only now comes back to macOS since they were the latest to diversify from one to multiple arches. (See my comment just above, as well as the note from your original PR post.)

So, the original hard-coding was not flexible enough, and thankfully we can do `process.arch` now. I have to guess/assume this wasn't available back in Node 0.11, and before Electron existed yet... So, maybe fudging the arch was important back when using Node in really non-standard ways??? Who knows?!

Also: By the way, this code is the only instance of `ATOM_ARCH` either in this repo or Pulsar core repo. So, I think it's just  a convenient way to make `ppm` rebuild for a different arch. Might as well leave it, but I would be surprised if there are no real users of that env var, and **the `getElectronArch()` function may be superfluous to simply using the `process.arch` method directly.** But again, doesn't seem like a big deal to have this code path, might be nice to use in a pinch, and any users who did need it would (hypothetically???) probably sorely miss it. So, **might as well keep it, I guess.**

savetheclocktower:
Yeah, I like the ability to specify `ATOM_ARCH` as an environment variable as an override just in case there is some weird scenario in which `process.arch` returns an unexpected value… or someone need to cross-compile.

DeeDeeG:
Very good to have this fix, thank you for the persistence in tracking this down!!

PR #146, opened 2024-12-01
Asyncify request

2colours:
This is basically https://github.com/pulsar-edit/ppm/pull/140 but now I didn't mess the branches up.

### DIFF
--- a/src/develop.js
+++ b/src/develop.js
@@ -43,25 +43,31 @@ cmd-shift-o to run the package out of the newly cloned repository.\
       return options.alias('h', 'help').describe('help', 'Print this usage message');
     }
 
-    async getRepositoryUrl(packageName) {
-      const requestSettings = {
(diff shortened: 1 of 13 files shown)


confused-Techie:
Unfortunately, GitHub isn't showing any diff on this PR.

Although the changes you intended in this PR are still available on #147 so I'll see what I can do about reviewing them there, and chime in back here if need be.

DeeDeeG:
Yeah, it looks like there is one commit here with what I presume is the full intended diff, and a second commit with the same message but reverting the changes of the first commit.

It may be possible to review this PR on the basis of the first commit, and to remember to delete the second commit before if/when it would be merged.

2colours:
I think since there has been (seemingly?) plenty of thought going into the other PR which does come with these changes, and not a lot of visible intent to fix/reuse this PR, perhaps it can be just closed at this point. The modifications in #147 still come in distinct commits so I don't think anything gets lost.

PR #147, opened 2024-12-01
Eliminate trivial underscore

2colours:
Now, only the overly custom part of underscore-plus remains: this config reading and writing that should rather get a separate module/package.

The implementations themselves could be further improved as better Set operations land, mostly in Node v22.

### DIFF
--- a/src/ci.js
+++ b/src/ci.js
@@ -2,7 +2,6 @@
 const path = require('path');
 const fs = require('./fs');
 const yargs = require('yargs');
-const _ = require('underscore-plus');
 
 const config = require('./apm');
 const Command = require('./command');
@@ -54,7 +53,11 @@ but cannot be used to install new packages or dependencies.\
 
     fs.makeTreeSync(this.atomDirectory);
 
-    const env = _.extend({}, process.env, {HOME: this.atomNodeDirectory, RUSTUP_HOME: config.getRustupHomeDirPath()});
+    const env = {
+      ...process.env, 
+      HOME: this.atomNodeDirectory,
+      RUSTUP_HOME: config.getRustupHomeDirPath()
+    };
     this.addBuildEnvVars(env);
 
     const installOptions = {env, streaming: options.argv.verbose};
--- a/src/command.js
+++ b/src/command.js
@@ -57,7 +57,7 @@ class Command {
   sanitizePackageNames(packageNames) {
     packageNames ??= [];
     packageNames = packageNames.map(packageName => packageName.trim());
-    return Array.from(new Set(packageNames)).filter(Boolean);
+    return Array.from(new Set(packageNames)).filter(Boolean); // Array of only unique truthy values
   }
 
   logSuccess() {
--- a/src/command.js
+++ b/src/command.js
@@ -1,7 +1,6 @@
 
 const child_process = require('child_process');
 const path = require('path');
-const _ = require('underscore-plus');
 const semver = require('semver');
 const config = require('./apm');
 const git = require('./git');
@@ -58,7 +57,7 @@ class Command {
   sanitizePackageNames(packageNames) {
     packageNames ??= [];
     packageNames = packageNames.map(packageName => packageName.trim());
-    return _.compact(_.uniq(packageNames));
+    return Array.from(new Set(packageNames)).filter(Boolean);
   }
 
   logSuccess() {
--- a/src/develop.js
+++ b/src/develop.js
@@ -82,7 +82,7 @@ cmd-shift-o to run the package out of the newly cloned repository.\
     installDependencies(packageDirectory, options) {
       process.chdir(packageDirectory);
 
-      return new Install().run({...options});
+      return new Install().run({...options}); // pass options shallow-copied - protects somewhat from mutations
     }
 
     linkPackage(packageDirectory, options) {
--- a/src/develop.js
+++ b/src/develop.js
@@ -2,7 +2,6 @@
 const fs = require('fs');
 const path = require('path');
 
-const _ = require('underscore-plus');
 const yargs = require('yargs');
 
 const config = require('./apm');
@@ -81,16 +80,17 @@ cmd-shift-o to run the package out of the newly cloned repository.\
     }
 
     installDependencies(packageDirectory, options) {
-        process.chdir(packageDirectory);
-        const installOptions = _.clone(options);
+      process.chdir(packageDirectory);
 
-        return new Install().run(installOptions);
+      return new Install().run({...options});
     }
 
     linkPackage(packageDirectory, options) {
-      const linkOptions = _.clone(options);
+      const linkOptions = {
+        ...options,
+        commandArgs: [packageDirectory, '--dev']
+      };
 
-      linkOptions.commandArgs = [packageDirectory, '--dev'];
       return new Link().run(linkOptions);
     }
 
--- a/src/develop.js
+++ b/src/develop.js
@@ -43,31 +43,25 @@ cmd-shift-o to run the package out of the newly cloned repository.\
       return options.alias('h', 'help').describe('help', 'Print this usage message');
     }
 
-    getRepositoryUrl(packageName) {
-      return new Promise((resolve, reject) => {
-        const requestSettings = {
-          url: `${config.getAtomPackagesUrl()}/${packageName}`,
-          json: true
-        };
-        return request.get(requestSettings, (error, response, body) => {
-          body ??= {};
-          if (error != null) {
-            return void reject(`Request for package information failed: ${error.message}`);
-          }
+    async getRepositoryUrl(packageName) {
+      const requestSettings = {
+        url: `${config.getAtomPackagesUrl()}/${packageName}`,
+        json: true
+      };
+      const response = await request.get(requestSettings).catch(error => Promise.reject(`Request for package information failed: ${error.message}`));
+      const body = response.body ?? {};
 
-          if (response.statusCode === 200) {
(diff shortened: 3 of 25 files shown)


2colours:
Gazing through the differences, this PR also seems to build upon #146 - that wasn't intentional but honestly, it's not "wrong enough" for me to try to go back. The content of that has been around for a long enough time, it's not going to change, and I think eventually it should get merged either way...

confused-Techie (changes requested):
I've reviewed your commit about removing outdated usage of `underscore-plus`.
Largely, I totally agree with the ways you've done so. Many of them being rather clever uses of built in features of the language.

But a recurring comment you'll find in my review is obscurity.

One obvious example:
```js
// === Original Code
return _.compact(_.uniq(packageNames));
```
Here, even if you aren't familiar with `underscore-plus` you can quickly assume what it's doing based on the function name `uniq` and can hopefully reasonably guess at what `compact` is doing.

But if we take what it's changed to in this PR:
```js
// === New Code
return Array.from(new Set(packageNames)).filter(Boolean);
```
While this code is generally faster by about `0.1ms` we've lost the clues at what's going on. Now requiring someone to be familiar enough with JavaScript to know that a `Set` only allows unique values in each collection. Sure `_.compact() => .filter()` is more obvious, but you get what I'm saying.

So in many places I've added requests for just some simple comments explaining the purpose of some code. Since we want to try and keep the barrier to entry low and not require too much knowledge of how everything works in JavaScript. But this really only applies to the more esoteric uses. Otherwise considering this is my only major point to make, the rest of this looks pretty rad. Thanks! I'll try and get to the async commit soon

src/search.js
```diff
@@ -44,23 +44,18 @@ Search for packages/themes.\
...
+      const body = response.body ?? {};
+
+      if (response.statusCode === 200) {
+        let packages = body.filter(pack => pack?.releases?.latest != null);
+        packages = packages.map(({readme, metadata, downloads, stargazers_count}) => _.extend({}, metadata, {readme, downloads, stargazers_count}));
+        packages = packages.filter(({name, version}) => !isDeprecatedPackage(name, version));
+        return packages;
+      }
+
+      const message = request.getErrorMessage(body, error);
```
confused-Techie, line 62: The second argument `error` here is never defined. Is it intended to be collected in a missing `try...catch` statement?

src/unstar.js
```diff
@@ -35,23 +35,15 @@ Run \`ppm stars\` to see all your starred packages.\
...
-          if (response.statusCode !== 204) {
-            this.logFailure();
-            const message = request.getErrorMessage(body, error);
-            return void reject(`Unstarring package failed: ${message}`);
-          }
+      const response = await request.del(requestSettings).catch(error => { this.logFailure(); throw error; });
+      const body = response.body ?? {};
+      if (response.statusCode !== 204) {
+        this.logFailure();
+        const message = request.getErrorMessage(body, error);
```
confused-Techie, line 42: We will want to make sure to have the second parameter be `null` here, as `error` doesn't exist

src/install.js
```diff
@@ -209,32 +209,26 @@ Run ppm -v after installing Git to see what version has been detected.\
...
-            return void reject(`Request for package information failed: ${message}`);
-          }
-          if (!body.releases.latest) {
-            return void reject(`No releases available for ${packageName}`);
-          }
-
-          resolve(body);
-        });
+      const response = await request.get(requestSettings).catch(error => {
+        const message = request.getErrorMessage(body, error);
```
confused-Techie, line 226: The first parameter here `body` doesn't exist, as we are only inside the `catch()` callback.

confused-Techie (changes requested):
Just reviewed your changes to async for requests.
Overall it seems solid, I just had to ignore the issues me precursor PR is trying to resolve, as those aren't on you.

Other than a few mistakes of leaving in variables that don't seem to exist, this seems rather solid and should function fine.
Ideally, once merged I can refactor my PR and add back the changes I'm trying to get for better error logging.

2colours:
> Here, even if you aren't familiar with `underscore-plus` you can quickly assume what it's doing based on the function name `uniq` and can hopefully reasonably guess at what `compact` is doing.
> 
> But if we take what it's changed to in this PR:
> 
> ```js
> // === New Code
> return Array.from(new Set(packageNames)).filter(Boolean);
> ```
> 
> While this code is generally faster by about `0.1ms` we've lost the clues at what's going on. Now requiring someone to be familiar enough with JavaScript to know that a `Set` only allows unique values in each collection.

I suppose we must be very different people because I disagree categorically. Of course the strawman to beat would be `compact` which I find totally opaque but I cannot think of a scenario when one might think, using any tech stack, that something called a "set" would allow duplicates, and furthermore that taking a list of values, turning it into a set, and then turning it back into an array could serve _any other purpose than making the elements unique_. I strongly believe this is the reason there isn't any other built-in for this; it's obvious enough and better yet, portable across programming languages and technologies. I would agree that the uniqueness is achieved by unfolding something that seems more like an implementation detail - but that could actually be a plus in a lot of cases and I still don't think it does any significant damage.

src/ci.js
```diff
@@ -54,7 +53,11 @@ but cannot be used to install new packages or dependencies.\
 
     fs.makeTreeSync(this.atomDirectory);
 
-    const env = _.extend({}, process.env, {HOME: this.atomNodeDirectory, RUSTUP_HOME: config.getRustupHomeDirPath()});
+    const env = {
+      ...process.env,
```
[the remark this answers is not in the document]
2colours: I also wanted to be sure and specifically checked MDN - it's guaranteed that what comes later overrides earlier values. I really share the mixed feelings here: if everybody knew this, probably we wouldn't fear it but still, it doesn't seem trivial...

src/command.js
```diff
@@ -58,7 +57,7 @@ class Command {
   sanitizePackageNames(packageNames) {
     packageNames ??= [];
     packageNames = packageNames.map(packageName => packageName.trim());
-    return _.compact(_.uniq(packageNames));
+    return Array.from(new Set(packageNames)).filter(Boolean);
```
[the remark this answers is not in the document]
2colours: It would be hard to object against a banal comment for some logic that will probably never change but still, I do think that this concern is overreaching. I specifically chose `.filter(Boolean)` because I felt that was the clearest way to do it - way clearer than the misleading `_.compact` itself - and I also think it's common knowledge that you turn something to a set and then back in order to eliminate duplicates. Of all the ppm code, I definitely wouldn't be the most concerned about this...

src/develop.js
```diff
@@ -81,16 +80,17 @@ cmd-shift-o to run the package out of the newly cloned repository.\
     }
 
     installDependencies(packageDirectory, options) {
-        process.chdir(packageDirectory);
-        const installOptions = _.clone(options);
+      process.chdir(packageDirectory);
 
-        return new Install().run(installOptions);
+      return new Install().run({...options});
```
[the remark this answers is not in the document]
2colours: Again, almost sure I checked all functions and what they do, and also, I don't think it's all that obscure; specifically, I feel that merely passing a shallow copy without saying anything about why that's necessary is way more obscure than the spread syntax way of creating a shallow copy.

src/disable.js
```diff
@@ -59,21 +59,21 @@ Disables the named package(s).\
...
+          packageNames = packageNames.filter(elem => installedPackageNames.includes(elem));
 
           if (packageNames.length === 0) {
             return "Please specify a package to disable"; //errors as return values atm
           }
 
           const keyPath = '*.core.disabledPackages';
           const disabledPackages = _.valueForKeyPath(settings, keyPath) ?? [];
-          const result = _.union(disabledPackages, packageNames);
+          const result = Array.from(new Set([...disabledPackages, ...packageNames]));
```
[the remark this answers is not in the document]
2colours: This is kind of a hack, after all; one that I probably found on SO. I don't know whose idea was to release Set in Javascript without adding set operations... they are coming with Node 22 iirc. I'm rather surprised I didn't outright add a TODO to replace these calls once the actual set operations land... it needs to be taken care of, one way or another.

savetheclocktower:
I tend to fall more on @2colours’s side of this disagreement, but some short explanatory comments don't seem like they're a hardship here.

src/search.js
2colours: I don't know, I didn't create `getErrorMessage` to begin with. 😅

src/develop.js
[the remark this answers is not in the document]
confused-Techie: I totally agree that doing a shallow copy with no comment originally is more obscure, because when I wrote that even I was wondering why they ever needed to do that.

But I'm sure you checked the functions and what they do because I couldn't find any differing behavior in your changes which is awesome.

src/search.js
confused-Techie: Well I just mean, your changes removing the definition of the variable `error`. So I was asking if you intended to have it defined elsewhere, but if not then we will want to instead do `request.getErrorMessage(body, null);`

src/install.js
2colours: I feel this was "too clever". I rather don't want to change the logic here at the moment, I will revert to the original way of producing the error message here.

2colours:
Alright, I addressed all the open issues. I think. Also, since the other PR is kinda spoiled and we are already addressing the request async topic here, I will just close that.

confused-Techie (approved):
@2colours Thanks for addressing my concerns even if they weren't totally necessary. I appreciate it, and lets get this one merged!

PR #148, opened 2024-12-31
Report errors from backend when a request fails (attempt 2)

confused-Techie:
After we merged #147 it created significant conflicts with my original PR (as it touched all the exact same places my PR did). So for the benefit of the contributor we merged that one first, and I reworked this one. 

If you'd like to read more about the original intentions of this PR (which are the same across the two attempts) feel free to look at #139 

---

A major pain point for community package authors is that nearly every error on the backend is returned with `Internal Server Error`. I ensured to create as helpful of error messages as possible to try to alleviate this, until I learned that PPM itself was discarding those error messages and just returning a status code.

There were already hints in the code at wanting to fix this via `request.getErrorMessage()` but it is severely under-utilized. So all this PR does is ensure we use it everywhere applicable. So that our users always are getting the best error message possible.

Again for a more in-depth explanation feel free to refer to #139


Daeraxa:
And another one that seems to have been missed - is this one still active/relevant and  looking to be merged?

confused-Techie:
> And another one that seems to have been missed - is this one still active/relevant and looking to be merged?

Thanks for finding this one, but yes I'd still love to get this one merged and get human readable error messages into PPM!

Not sure if we want to run tests again for sanity, but if we do, we likely should wait on CI fixes to get into `master` from #161 and then we can update from `master` to rerun specs here

Daeraxa:
161 has landed at least so happy to have a proper look at this if brought up to speed with `master`

confused-Techie:
@Daeraxa I've updated this PR with changes from `master` for whenever you've got time

Daeraxa:
Did a bit of quick testing, wasn't able to get anything with a different error response displayed from some basic attempts from doing silly things but the changes don't look big enough to cause any issues either. Unofficial approval only im afraid.

Daeraxa (approved):

confused-Techie:
Thanks a ton for the review @Daeraxa! But yeah for my testing I was targeting a specific error message to test with, so far to even mangle the data of my package on the database. But with these changes, any of the error messages we test for within [the full HTTP tests on the backend](https://github.com/pulsar-edit/package-backend/tree/main/tests/full) should now be fully accessible. 

But I'll go ahead and merge!

PR #149, opened 2025-01-07
Remove dependency on "request" package

2colours:
It was only used in the download-node script - now that also uses superagent instead.

I tested the download locally and it did succeed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,6 @@
     "open": "7.3.0",
     "plist": "3",
     "read": "~1.0.7",
-    "request": "^2.88.2",
     "season": "^6.0.2",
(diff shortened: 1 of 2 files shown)


confused-Techie (approved):
The diff here looks to have identical meaning. And I've tested to ensure `NodeJS` is still being installed successfully, and contains the same amount of data between installs.

Great work on this final user of `request`, I'll be happy to see us finally get rid of it!

DeeDeeG (approved):
Looks good to me as well. 👍 

CI is passing, with a successful rebuild of `ppm`'s own dependencies against the downloaded Node, so the download script must be working fine. Works for me locally as well.

With two Approves, I don't see why not to merge, so I'll go ahead and merge this.

Thanks again.

PR #150, opened 2025-02-20
Update to read v3

2colours:
Read v4 declares a dependency on Node v18 at least and the bundled Node version is v16.0.0 still. The modifications aren't big but that upgrade would be the way even further. Other than that, this package didn't change a lot. It's supposed to "just work" if one exports the `read` function appropriately (not as a "default" import), with an async interface that we pretty much already had because of the promptX kind of methods.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "npm": "https://github.com/pulsar-edit/npm-cli/releases/download/v6.14.19-pulsar2/npm-6.14.19-pulsar2.tgz",
     "open": "7.3.0",
     "plist": "3",
-    "read": "~1.0.7",
+    "read": "^3",
     "season": "^6.0.2",
     "second-mate": "https://github.com/pulsar-edit/second-mate.git#9686771",
     "semver": "^7.3.4",
--- a/src/login.js
+++ b/src/login.js
@@ -1,6 +1,6 @@
 
 const yargs = require('yargs');
-const read = require('read');
+const { read } = require('read');
 const open = require('open');
 
 const auth = require('./auth');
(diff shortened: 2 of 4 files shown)


mauricioszabo (approved):
:clap:

DeeDeeG:
This removes error handling. `read()` can possibly error, as its usage documentation suggests to put it in a try/catch: https://github.com/npm/read/blob/main/README.md#usage

```js
const { read } = require('read')
// or with ESM: import { read } from 'read'
try {
  const result = await read(options)
} catch (er) {
  console.error(er)
}
```

With a default "yes," I think its safer to error out than continue if `read()` didn't work.

Per StackOverflow (https://stackoverflow.com/questions/42453683/how-to-reject-in-async-await-syntax), we can `try/catch` and `throw` the error to make the async function we are in return a rejected promise with the error from `read()`... Or add a `.catch( . . . )` to the `read()` call like:

```js
      let answer = await read({prompt: 'Would you like to install these updates? (yes)', edit: true}).catch(error => throw(error));
```

Something like that.

Perhaps I can investigate the right way to restore error handing myself, but you are also free to attempt it.

DeeDeeG:
My suggestion above was discussed on Discord, and it was pointed out to behave equivalently with or without try/catch. I can confirm that is true in my own testing now.

There is minimal logging of the error in this situation regardless, it looks like this if throwing just before `read()` would have been called in `upgrade.js` on this branch, or rejecting the promise just before `read()` would have been called on `master` branch:

```console
% ./bin/ppm upgrade deedeeg-new-package-1
Package Updates Available (1)
└── deedeeg-new-package-1 0.0.1 -> 0.0.9

some error string (this is red)
```

**Since it's the same as before this PR, there is no issue.** Thanks for bearing with me while I test things locally.

DeeDeeG (approved):
Code looks reasonable, moreso after discussion on Discord. I confirmed it behaves equivalently, and it is functional (upgrading my package on the CLI works) just the same as `master` branch.

So, Approved from me as well. Merging. Thanks for the dependency bumps lately.

PR #152, opened 2025-07-27
Add typescript template

savetheclocktower:
Marking this as a draft for now because further testing needs to be done, but this is part of the work needed for [pulsar-edit#1272](https://github.com/pulsar-edit/pulsar/issues/1272).

This adds the ability to generate a TypeScript template by running `ppm init --package <package-name> --syntax typescript`. We already allowed people to choose between JavaScript and CoffeeScript when generating a basic package template; this just adds TypeScript as another option.

This TypeScript template does all of what's described in  [pulsar-edit#1272](https://github.com/pulsar-edit/pulsar/issues/1272):

* Prefers CommonJS versions of dependencies when available, but will transpile ESM to CommonJS.
* Generates source maps.
* Allows direct importing of JSON files.

It also has our fork of `@types/atom` present in `devDependencies` such that type annotations will be present in one's IDE for the entire `atom` API. (Install [pulsar-ide-typescript](https://web.pulsar-edit.dev/packages/pulsar-ide-typescript) to enjoy this feature, of course.)

This needs to be tested in practice before landing. I used this exact structure (albeit constructed by hand rather than generated) for [pulsar-hover](https://web.pulsar-edit.dev/packages/pulsar-hover) and it served me well. I've also verified that that package works equally well in mainline Pulsar as it does in PulsarNext.

What's left to test is running `ppm init -p foo -s typescript` and asserting that it generates a package that “works” out of the box (doesn't fail with syntax or build or load errors). This might be tricky; unlike other template types, this one's entry point is a generated file, so it won't be present after generation until the user runs `npm run install` and `npm run build`. The answer might be to make a stub `lib/index.js` as part of the template so that we can catch this and tell the user what to do.

---

I also made very slight changes to how `ppm init` works:

* When `-s`/`--syntax` was omitted, we were assuming the user wanted a CoffeeScript package. This absolutely should not be the default any longer. We can keep supporting it without much hassle, but we should stop advertising it in the help text, too. When `-s` is omitted, we now assume JavaScript. The specs have been updated accordingly.
* The `result` parameter in `generateFromTemplate` was a classic relic of decaffeinated code. After verifying we weren't actually using that return value for anything, I removed it.


src/init.js
```diff
@@ -122,30 +136,31 @@ on the option selected.\
...
           contents = this.replacePackageNamePlaceholders(contents, packageName);
           contents = this.replacePackageAuthorPlaceholders(contents, packageAuthor);
           contents = this.replaceCurrentYearPlaceholders(contents);
-          result.push(fs.writeFileSync(sourcePath, contents));
-        } else {
-          result.push(undefined);
+          fs.writeFileSync(sourcePath, contents);
         }
       }
-      return result;
```
2colours, line 148: So you did confirm that nobody uses this return value, right?

savetheclocktower: Yeah, the only calls to `generateFromTemplate` come from within this file, and none of them do anything with the return value.

savetheclocktower:
I think I _meant_ to make this a draft, but can't find any evidence that it ever _was_ a draft. However, now that I've done the manual testing I promised, pretend I just took this out of draft!

savetheclocktower:
As usual, the CI needs a major medical intervention because we committed the crime of relying on something.

DeeDeeG:
I did a CI fixup in #154. If you like, I can rebase this branch on top of that, or fix the merge conflict.

I'm not super involved with TypeScript, but I'm inclined to believe you that this is a good way to set up [a template for] new TypeScript packages. (And it can always be improved later on down the line.) CI passes with updated tests, so that's nice.

DeeDeeG:
> Yeah, feel free to update this PR!

It is done. 👍

PR #153, opened 2025-09-27
Update bundled Node from v16.0.0 to v18.20.8

DeeDeeG:
Node 16 has been out of support (EOL) for a long time now.

Node 18 has only somewhat more recently gone out of support.

We should update to something newer, for the usual performance and security reasons.

I don't think ppm compiles against Node 20, though, so this is the furthest trivial upgrade we can do, as far as I know.

Better than nothing.


savetheclocktower:
We might want to go further and start explicitly keeping `ppm`’s Node and Pulsar’s Node in sync — manually to start, then eventually forcing `ppm` to use the actual Pulsar executable as its Node (via `ELECTRON_RUN_AS_NODE=1`).

When testing the new `terminal` package on Windows, I ran `ppm install pulsar-edit/terminal` and immediately ran into trouble. It tried to install using `ppm`’s version of Node, but a build script somewhere was relying on `Array.prototype.at`, which wasn't introduced to Node until v16.6. (EDIT: The culprit is `rollup` version 4, which [supports Node 18 and above](https://github.com/rollup/rollup/blob/master/package.json).)

This PR would've solved this specific case, but any version discrepancy between `ppm` and Pulsar could cause similar problems. This is why `node` and `npm` are themselves so coupled together.

If it's not a gigantic pain in the ass to bump `ppm`’s bundled Node to `20.11.1` before we ship the Electron 30 upgrade, we should think about it. If it would be a pain, then we can hold off on it — but we should land this PR at a minimum.

That said, I'm happy to learn more about this infrastructure so that I can help more with this effort in the future. This shouldn't be just one person's job!

savetheclocktower:
OK, the good news is that that problem had an easy fix: I was being silly and not giving this checked-out `ppm` repo any way of finding the correct Pulsar installation so it could extract Electron version info. Once I did

```
ATOM_RESOURCE_PATH="/path/to/pulsar-on-electron-30" bin/ppm install pulsar-edit/terminal
```

everything seemed to work great.

DeeDeeG:
Can think about revisiting this if Node 18 ever becomes desirable, such as for an alternate branch of `ppm`, but I don't have perfect foresight to know if that will ever be the case.

The main obvious thing is that #156 is great for Electron 30 stuff, and hopefully in any other context as well. But we shall find out, I guess.

PR #155, opened 2025-11-03
Fix misnamed variable in `rebuild` command

savetheclocktower:
This is a fix for [the bug I described here](https://github.com/pulsar-edit/pulsar/pull/1367#issuecomment-3479172584). This is an artifact of an earlier rewrite and updated exception handling; `stderr` now exists in another scope and is thrown, so we should catch it and treat `error` like a string.

I also noticed that, when `stderr` is empty, we just reject with an empty string in `forkNpmRebuild`. In that case, I figure a generic message with the failure code is better than an empty string, so I changed it. Happy to take that out if it bothers anyone.

I've not made much progress in figuring out [how to hit this particular code path on demand](https://github.com/pulsar-edit/pulsar/pull/1367#issuecomment-3479415089), though! If I find a way, I'll update the PR.

### DIFF
--- a/src/rebuild.js
+++ b/src/rebuild.js
@@ -52,7 +52,7 @@ All the modules will be rebuilt if no module names are specified.\
       return new Promise((resolve, reject) =>
         void this.fork(this.atomNpmPath, rebuildArgs, {env}, (code, stderr) => {
           if (code !== 0) {
-            reject(stderr ?? `Unknown error: code ${code}`);
+            reject(stderr ?? `Unknown error while invoking npm: code ${code}`);
             return;
           }
 


DeeDeeG:
I can confirm the `stderr` should be `error` now.

(It used to be `stderr` at this point in the history: https://github.com/pulsar-edit/ppm/blame/2cb92e55808c6c6f2011324d4b229be3035a552f/src/rebuild.js#L72-L74 but a refactor changed the variable in question's name to `error` without changing what's logged.)

DeeDeeG (approved):
Looks good to me!!!

Thank you for this, when I read the description of the problem I was ready to investigate and open a PR, but lo and behold, here is such a PR. Brilliant.

Optional code suggestion below, otherwise, feel free to land this. 👍

PR #156, opened 2025-11-12
Upgrade Node to version 20.11.1

savetheclocktower:
Inspired by #153, I decided to tilt at this windmill, and in particular to upgrade the specs to the latest version of Jasmine.

This was a chore, but a straightforward one, and it let me shed a lot of boilerplate. And once I bumped the ancillary fixture artifacts to agree with Node 20.11.1, the last few tests started passing! (For me locally, at least. We'll see how humbling the CI run is.)

### DIFF
--- a/docs/bumping-the-node-version.md
+++ b/docs/bumping-the-node-version.md
@@ -1,0 +1,38 @@
+# Bumping Node
+
+Ideally, the version of Node used by `ppm` should be kept in sync with the version of Node used by the Electron version of the corresponding release of Pulsar.
+
+For example: at time of writing, Pulsar uses Electron 30.0.9, which corresponds to Node 20.11.1. So `ppm` should also point to Node 20.11.1.
+
+You should know **both** the Node version **and** the Electron version you’re targeting before you begin; you’ll use both values.
+
+Here’s how it works:
+
+## Run the specs
+
+**Before you change anything**, run `bin/npm test` and ensure the test suite is green. If it isn’t, don’t proceed — unless you have reason to believe that the Node upgrade itself is what will fix the failing specs.
+
+## Change `BUNDLED_NODE_VERSION`
+
+The contents of this file govern which version of Node is downloaded when `ppm` is installed. Make sure this corresponds to the new **Node** version that `ppm` will use, since this will be downloaded directly from the Node servers when `ppm` is built.
+
+
+## Change the fixture files
+
+The `spec/fixtures/node-dist` directory contains files that are needed for specs that test `ppm`’s ability to compile dependencies with native modules.
+
+We use _Electron’s_ artifacts here because they’re smaller than Node’s. So even though we used the Node version number above, we’ll need `spec/config.json` to point to the corresponding Electron version number, not the Node version number!
+
+Luckily, there’s a script to automate all of this. From the project root, and assuming you want to grab the artifacts for Electron 30.5.1, run:
+
+```sh
+node ./scripts/update-spec-fixtures.js 30.5.1
+```
+
+This will construct the correct URLs, download these files to `spec/fixtures/node-dist` under the names that the specs expect, and finally update `spec/config.json` once the downloads are finished.
+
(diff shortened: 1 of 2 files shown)


DeeDeeG (commented):
**Essential info is bolded,** the rest is fluff!

(This is a not-very-obvious thing, I do not expect this to be apparent without a mention. So here I am pointing it out...!)

**Please consider using Electron headers rather than Node headers for the spec fixtures.** (They are smaller is the main motivation. Also, more-closely tests with a like-for-like header version as we anticipate using in Pulsar core, to the extent that ends up mattering.) The electron header files can be fetched from electron.org servers by poking URLs with the following scheme: **e.g.** https://artifacts.electronjs.org/headers/dist/v17.0.0/node.lib; **https://artifacts.electronjs.org/headers/dist/v30.0.9/node-v30.0.9-headers.tar.gz**

(See: https://www.electronjs.org/blog/s3-bucket-change for the lore behind these URLs if curious.)

The filename scheme is inherited from Node stuff for tooling compatibility, so the immediate impression is, "of course these are Node headers and libs, etc.!"

(Story time: I only remembered the Electron headers existed and could be used here after committing some rather large Node 10 headers once... their bloat haunts the repo now when `git clone`ing...)

(TANGENT: This also makes me realize we should be looking into getting Electron 30 in core stuff to be 30.5.1, not 30.0.9.

I found this tool today to check the release history: https://releases.electronjs.org/release?major=v30 and it made me realize 30.0.9 isn't the most-recent Electron 0 anymore!

But that's there not here!)

DeeDeeG:
It's kind of fiddly to update the spec files, and I've refreshed my memory of how to do it (via a trip down memory lane), so I can get those files updated if you like. But I also respect the "passing down of knowledge" aspect if you prefer to do it yourself and learn how hands-on. (So I will assume you want to do it as indicated above, but feel free to request me to do it if it's preferable to have one more fiddly task off of your plate.)

EDIT: I note you've already gotten it right once already, given the tests aren't exploding! Well done there! I recall it being a bit confusing and reverse-engineering-ish when I first did it.

DeeDeeG (commented):
I have some minor nit-picks:

- I'm not sure how I feel about the waterfall/pyramid-of-doom indentation for the `app.get()` calls for `express` in various spec files for readability, but this is such a minor point I'm not really gonna be too bothered if it stays.
- Some oddness ess and churn in `yarn.lock` such as `registry.npmjs.org` instead of `registry.yarnpkg.org` ... I restored `yarn.lock` from `master` and reverted the `second-mate` change in `package.json` and re-ran `yarn install` for a much smaller and more normal-looking diff. I can push those two changed files if you like, but for cleaner history I might prefer the churn be rebased out entirely. But this isn't such a sticking point that I can't hear a counter argument.

But overall looking good, I would save my writeup of what I like for the approve, but it is pretty nice IMO!

savetheclocktower:
I certainly think the `yarn.lock` nitpick is worth adopting. I don't know what caused that much churn in the lockfile, but I'm definitely in favor of smaller diffs. Please do push the lockfile changes, and rewrite history however you like!

DeeDeeG (approved):
Lots less boilerplate in the specs, a more modernized Node version (and matching the anticipated Electron 30 bump in core), a more modernized Jasmine, a script to automate getting newer spec fixtures. All pretty cool, thank you for these things!

The diff to specs is rather small outside of the boilerplate reduction, some promisification and usage of `async`/`await`, and whitespace/indentation stuff, which is great, IMO.

Note: I think the old local-jasmine stuff was a workaround for some jasmine-focused funkiness, if I'm reading https://github.com/pulsar-edit/ppm/pull/87 correctly. So, hopefully moving away from jasmine-focused entirely is for the best. It was always a bit fiddly to have a patched Jasmine, I reckon. Probably for the best to move past that. And for a much newer Jasmine, so, yeah.

As mentioned above, I tested locally and tests passed there as well as them passing in CI.

LGTM, thanks once again, this seems like a bigger change to get done, but pretty smooth to review considering.

PR #161, opened 2025-11-23
Add “transpiled JavaScript” template

savetheclocktower:
This PR will be very similar to #152. I probably should've done both at once.

A Pulsar user writing a package for the first time in the year 2025 will discover that some of the dependencies they might want to use are ESM-only. Over time, this will become an even more common method of distribution. This is a problem for Pulsar because the renderer process is CJS-only, at least if you want to be able to use dependencies from the Node universe. This is a rigid requirement and Electron is unlikely to be able to change it in the near future.

This won't affect _some_ kinds of packages. And, indeed, if you want to solve the ESM-only problem without transpiling your code, you can probably find an older version of the same dependency that had a CJS bundle, or find another package that does a similar thing and can serve up CJS. (Core packages can use these tactics to stave off transpilation for now.)

But it's better if users don't have to care about this! Hence this PR adds a new package template that sets them up with something that will automatically transpile ESM dependencies to CJS, where possible. (It's possible unless the ESM module uses bleeding-edge features like top-level `await`, but this is pretty rare.)

The TypeScript template in #152 already had this toolchain. This is the same thing, but for people who prefer to write in JavaScript.

This template therefore omits the TypeScript dependencies from `package.json`… but it keeps the `@types/atom` dependency (the one that points to our own `@pulsar-edit/types` package) because those types are handy to have even for JS-only projects if the user has `pulsar-ide-typescript` installed. (Also, now that our types package is published to NPM, I've updated the `package.json` templates for both transpiled-JS and TypeScript to reference the NPM version of the library.)

## Testing

Specs are updated. But for manual testing, you can run something like

```sh
bin/npm init --package some-package --syntax javascript-transpiled
```

to generate a JS package and opt into the “transpiled” syntax. (I decided to add `javascript-transpiled` as another valid option for `--syntax` rather than introduce a one-off flag that has meaning only when other specific options are present.) Once you've generated the package, `cd` into the new directory, run `ppm link --dev .` to create a symlink in to `~/.pulsar/dev/packages`, then run `pulsar --dev .` to open a new dev-mode window for the package.

You should be able to invoke **(name of your package): Toggle** in the command palette and trigger the “you must transpile this package first!” notification. You should also be able to run `npm install` and then `npm run build` from the terminal, then reload your Pulsar window… and now **(name of your package): Toggle** should do the standard thing (create a new panel and put some silly text in it).

### DIFF
--- a/src/init.js
+++ b/src/init.js
@@ -68,7 +68,7 @@ on the option selected.\
           // Expose the error value as a value for now.
           return `You must specify one of ${this.supportedSyntaxes.join(', ')} after the --syntax argument`;
         }
-        templatePath = this.getTemplatePath(options.argv, syntax);
+        templatePath = this.getTemplatePath(options.argv, `package-${syntax}`);
         this.generateFromTemplate(packagePath, templatePath);
         return;
       }
@@ -192,15 +192,14 @@ on the option selected.\
       return string.replace('__current_year__', new Date().getFullYear());
     }
 
-    getTemplatePath(argv, syntax) {
+    getTemplatePath(argv, templateType) {
       if (argv.template != null) {
         return path.resolve(argv.template);
       }
-      let templateName = `package-${syntax}`;
-      if (syntax === 'javascript' && argv.transpiled) {
-        templateName = 'package-javascript-transpiled';
+      if (templateType === 'package-javascript' && argv.transpiled) {
+        templateType = 'package-javascript-transpiled';
       }
-      return path.resolve(__dirname, '..', 'templates', templateName);
+      return path.resolve(__dirname, '..', 'templates', templateType);
     }
 
     dasherize(string) {
--- a/src/init.js
+++ b/src/init.js
@@ -25,7 +25,7 @@ class Init extends Command {
       options.usage(`\
 Usage:
   ppm init -p <package-name>
-  ppm init -p <package-name> --syntax <javascript-or-typescript>
+  ppm init -p <package-name> --syntax <javascript-or-typescript> [--transpiled]
   ppm init -p <package-name> -c ~/Downloads/r.tmbundle
   ppm init -p <package-name> -c https://github.com/textmate/r.tmbundle
   ppm init -p <package-name> --template /path/to/your/package/template
@@ -47,6 +47,7 @@ on the option selected.\
       options.alias('l', 'language').string('language').describe('language', 'Generates a basic language package');
       options.alias('c', 'convert').string('convert').describe('convert', 'Path or URL to TextMate bundle/theme to convert');
       options.alias('h', 'help').describe('help', 'Print this usage message');
+      options.describe('transpiled', 'Choose the "transpiled" flavor of JavaScript project (valid only when --syntax is javascript)');
       return options.string('template').describe('template', 'Path to the package or theme template');
     }
 
@@ -67,7 +68,7 @@ on the option selected.\
           // Expose the error value as a value for now.
           return `You must specify one of ${this.supportedSyntaxes.join(', ')} after the --syntax argument`;
         }
-        templatePath = this.getTemplatePath(options.argv, `package-${syntax}`);
+        templatePath = this.getTemplatePath(options.argv, syntax);
         this.generateFromTemplate(packagePath, templatePath);
         return;
       }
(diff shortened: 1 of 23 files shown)


savetheclocktower:
Bump! Pinging @Daeraxa in case they'd like to do an informal review (but feel no obligation).

src/init.js
confused-Techie, line 1: Unless I'm misunderstanding the usage of the `--transpiled` flag, seems there's a couple leftover places in this file where we still try to do it.

* In the usage message `[--transpiled]`
* In description text `options.describe('transpiled'...`
* In `getTemplatePath` `&& argv.transpiled`

confused-Techie (changes requested):
The code here overall looks super solid.

I'd like to test personally, but with the tests you've written (and checking for specific files like a rollup) I feel like we can be pretty confident in this one functioning.

Although it seems like to no fault of your own the Windows tests broke. Seems like we can't find our version of Visual Studio code, so we might wanna add a fix for that here, which I'm assuming can match a manual install of an older version in the workflow file, like I'm pretty sure we do in Pulsar.

Otherwise I left a comment on one file that still seemed to assume we require the `transpile` flag for this new template, which it sounds like you're trying to do away with.

src/init.js
savetheclocktower: Whoops! I had made those change but left them unstaged. Just pushed another commit!

savetheclocktower:
Also just pushed a workflow file update pinning to `windows-2022` (as we've done elsewhere) and shifting the matrix forward to include Node 20. (We certainly don't need to be testing 14 and 16 anymore, and even 18 is probably unnecessary.)

templates/package-javascript-transpiled/spec/__package-name__-spec.js.template
```diff
@@ -0,0 +1,73 @@
+'use babel';
+
+import __PackageName__ from '../lib/index';
+
+// Use the command `window:run-package-specs` (cmd-alt-ctrl-p) to run specs.
```
Daeraxa: Only a small thing but as this is a macOS specific keybind should this provide the windows/linux binding as well or just remove it entirely?

templates/package-javascript-transpiled/styles/__package-name__.less.template
```diff
@@ -0,0 +1,8 @@
+// The ui-variables file is provided by base themes provided by Atom.
```
Daeraxa: I know we have a lot of these leftovers still but if this is just a copy of the original text of this file could we update this from Atom to Pulsar?

src/init.js
```diff
@@ -25,7 +25,7 @@ class Init extends Command {
       options.usage(`\
 Usage:
   ppm init -p <package-name>
-  ppm init -p <package-name> --syntax <javascript-or-typescript>
+  ppm init -p <package-name> --syntax <javascript,typescript,javascript-transpiled>
```
Daeraxa: Nitpicky but I think I mostly see this as separated by `|` in other help pages? Not a major thing either way.

templates/package-javascript-transpiled/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "__package-name__",
+  "type": "module",
+  "source": "./src/index.ts",
```
Daeraxa: should this be `index.js`?

templates/package-javascript-transpiled/lib/index.cjs
```diff
@@ -0,0 +1,11 @@
+const description = `After installing your dependencies with \`npm install\`, you must run \`npm run build\` or \`npm run watch\` from the root in order to compile your project — at which point you will no longer see this warning.
+
+If you're stuck, follow the directions in your new package's \`README.md\`.
+`;
+s
```
Daeraxa: A typo or just me missing something?

Daeraxa (commented):
Tried it out and it does what it says just fine. I've made some comments, some might be just nothing but at least a couple I think might need looking at.

templates/package-javascript-transpiled/package.json
savetheclocktower: Yes (or perhaps `./src/index.cjs`). I'll fix it.

templates/package-javascript-transpiled/lib/index.cjs
savetheclocktower: No, it's a typo, and a strange one. Will fix!

templates/package-javascript-transpiled/package.json
savetheclocktower: Actually, `source` is meant to point to the actual entry point of the source code. It isn't used by Node or `npm`, but it is used by tooling like Parcel, which maybe explains why I threw it in here in the first place. But I might just remove it — it serves no purpose with this template's Rollup-based bundling!

src/init.js
confused-Techie: Thanks for addressing the feedback!

confused-Techie (approved):
With all of my feedback addressed, as well as @Daeraxa's feedback, and with her actually trying the command directly, I think that should cover just about everything and we should be good to go!

EDIT::

Also thanks for fixing the CI here, I know it's entirely unrelated to your PR, but glad we could get it in somewhere we were already actively reviewing

PR #164, opened 2025-12-29
Eliminate mv, ncp, wrench dependencies

### LINKED ISSUE #157: Replace `wrench` with `fs-extra`/`klaw`
### Have you checked for existing feature requests?

- [x] Completed

### Summary

`ppm` uses [`wrench`](https://www.npmjs.com/package/wrench) in a number of places to handle filesystem tasks. But `wrench` is deprecated and the author recommends moving to [`fs-extra`](https://www.npmjs.com/package/fs-extra).

In fact, this could be done in concert with other optimization tasks around `fs`. There are needless usages of synchronous `fs` methods scattered around the codebase; my guess is that this is a relic from a time when it was more of a headache to use the asynchronous `fs` methods because of the callback pattern.

I mention `klaw` because we have one instance of a call to `wrench.readDirSyncRecursive`, and `fs-extra` no longer implements the equivalent of that function, having delegated it to the `klaw` package.

### What benefits does this feature provide?

It is good not to rely on deprecated packages. `fs-extra` is ridiculously popular and heavily used in the Node ecosystem.

### Any alternatives?

We could attempt to use `fs/promises` in the Node stdlib directly. I know that the `fs` library has evolved quite a bit toward solving the problems that necessitated the “enhancer” libraries like `fs-extra` and Atom’s own `fs-plus`. But I don't know whether `fs/promises` would check all our boxes on its own.

### Other examples:

_No response_

2colours:
Resolves #157. Comments might be worth reading.

This is meant to be a simple and incremental refactor with relatively meaningful commits. All spectests pass after the modifications, and 3 direct dependencies have been eliminated in favor of Node's built-in `fs/promises` API.

As I mentioned before, `fsPromises.cp` is an experimental utility in Node 20 - however, I see no reason to come up with an intermediary solution when Node 20 itself is basically "oldstable" now and Node 22 promoted the same API to stable. I wouldn't expect any sort of trouble but in any case, this clearly seems to be a straight path.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,8 +29,6 @@
     "git-utils": "^5.7.2",
     "hosted-git-info": "^3.0.7",
     "keytar": "^7.7.0",
-    "mv": "2.1.1",
-    "ncp": "~2.0.0",
     "npm": "https://github.com/pulsar-edit/npm-cli/releases/download/v6.14.19-pulsar2/npm-6.14.19-pulsar2.tgz",
     "open": "7.3.0",
     "plist": "3",
@@ -44,7 +42,6 @@
     "temp": "^0.9.4",
     "underscore-plus": "1.x",
     "wordwrap": "1.0.0",
-    "wrench": "~1.5.1",
     "yargs": "^3.32.0"
   },
   "devDependencies": {
--- a/script/download-node.js
+++ b/script/download-node.js
@@ -1,5 +1,5 @@
 const fs = require('fs');
-const mv = require('mv');
+const fsPromises = require('fs/promises');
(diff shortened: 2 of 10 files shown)


confused-Techie (approved):
Honestly this PR is simple enough, with lots of usage copied exactly from where it comes from, just using way less deps.

Love to see us slimming down the binary here, and especially love to see all the tests passing.

I'd say this one's good to go and touches on a grip I have in tracking down filesystem related errors where I used to have to dig several libraries deep to see what was happening.

Lets get it merged! (Especially before my rather large PR gets merged in the near future)

PR #167, opened 2026-02-19
Bump `tar` 

confused-Techie:
Replaces #166


Daeraxa:
This seems to have fallen between the cracks - is there a good reason to not approve this and bump `tar`?

confused-Techie:
> This seems to have fallen between the cracks - is there a good reason to not approve this and bump `tar`?

Not that I can think of, except CI isn't happy. We will likely want to rerun tests with fixed CI coming from another PR. And if tests are happy I can't think of any reason not to bump it

PR #168, opened 2026-02-22
Bump bundled Node to 20.16.0, to sync with Electron 30.5.1 in Pulsar core repo

DeeDeeG:
Bumping the bundled Node version to 20.16.0, to match Electron 30.5.1, which we should be able to update to in Pulsar core repo shortly.

Also bump our test fixture Electron headers to be 30.5.1, because why not? (I don't think this is strictly needed, and I might be inclined to do it once per major, but the files aren't _too_ massive. So, not toooooo much harm in it.)

Minorly enhanced the related documentation markdown as well.

[ Note: Opening this as draft, since `ppm` should pretty much follow Pulsar on these migrations, not the other way around. I think we can land this in sync just after the core repo PR lands, ideally.  See: https://github.com/pulsar-edit/pulsar/pull/1452, which I'll probably convert from draft to "ready for review" as it's going pretty well so far. ]


savetheclocktower:
> I kinda have it set up that way regardless -- I already bumped `ppm` submodule in the core repo PR to point to this branch tip's non-`master` commit of `ppm` repo.

Ah, so the submodule would not be affected by the merging of this PR to master. Clever!

As you were.

DeeDeeG:
Okay, the Electron bump at core repo landed (https://github.com/pulsar-edit/pulsar/pull/1452), so this present PR is also ready to land!

Reviews welcome and appreciated!

savetheclocktower (approved):
Looks great!

DeeDeeG:
Thank you! Merging.

PR #176, opened 2026-05-26
Fix publishing for packages with git+ protocol repository URLs

idleberg:
Tools such as `publint` encourage the usage of `git+https://` URLs in `repository.url`. However, `ppm` doesn't like this change. Similar to stripping out the `.git` extension, this PR strips `git+` from the protocol.

### DIFF
--- a/spec/packages-spec.js
+++ b/spec/packages-spec.js
@@ -1,5 +1,44 @@
 const Packages = require('../src/packages');
 
+describe('getRepository', () => {
+  it('returns null for missing metadata', () => {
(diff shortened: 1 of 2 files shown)


idleberg:
Wait a minute, I already fixed that in https://github.com/pulsar-edit/ppm/pull/173, seems there hasn't been a new version of `ppm` released since then 😅

savetheclocktower:
Created https://github.com/pulsar-edit/pulsar/issues/1578 to make sure we bump `ppm` before the next release. :)

PR #177, opened 2026-06-22
Chore: Update ppm licence file

Daeraxa:
I have no idea how this escaped our notice all this time... Just copies the same licence from the pulsar repo but with the GH copyright copied from the original apm repo.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,20 +1,22 @@
-Copyright (c) 2013 GitHub Inc.
+MIT License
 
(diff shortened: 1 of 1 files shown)


confused-Techie (approved):
Great catch! We know the failure of windows CI already has a fix in #161 so we shouldn't have to worry about that here. Lets get it merged!

Daeraxa:
I don't have merge permissions on here so ill have to rely on somebody :)

